PR #47, opened 2023-11-02
WIP: Adds Replicate LLM

dvsrepo:
Current implementation supports `public` endpoints (like the one used in the example). I couldn't find a way to not get a stream.

For `deployments` I have another implementation that works, but not with public deployments:

```python
    self.deployment = replicate.deployments.get(endpoint_url)
    def _text_generation_with_backoff(self, prompt: str) -> Any:
        prediction = self.deployment.predictions.create(
            input={
                "prompt": prompt,
                "max_new_tokens": self.max_new_tokens,
                "temperature": self.temperature,
            },
        )
        prediction.wait()
        return prediction.output
```

I think we should support both public (the easiest for newcomers) and private deployments.

### DIFF
--- a/examples/zephyr-replicate.py
+++ b/examples/zephyr-replicate.py
@@ -1,0 +1,41 @@
+import os
+
+from datasets import load_dataset
+from ultralabel.llm.openai_ import OpenAILLM
+from ultralabel.llm.replicate import ReplicateLLM
+from ultralabel.pipeline import Pipeline
+from ultralabel.tasks.preference.ultrafeedback import MultiRatingTask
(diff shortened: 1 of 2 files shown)


alvarobartt:
Hi here! I'm tackling this now, and I'm afraid the output will most likely be an iterator so we'll need to wrap it around with `list(output)`, see the lines below from `replicate-python`

https://github.com/replicate/replicate-python/blob/cd09db0691a96c3c299e1f9ac374d420bb4adbdd/replicate/run.py#L46-L47

But w.r.t. the stream, we can just pass the flag `stream=False`, I'll open a PR later today so that you can review!

PR #116, opened 2023-11-27
Fixes None issues in to_argilla ultrajudge

dvsrepo:


alvarobartt:
To be closed in favour of #119, but needs confirmation first, @dvsrepo could you send me privately the example you ran that was leading to this issue so that I can reproduce and ensure that the refactored approach works well? Thanks in advance!

PR #139, opened 2023-12-04
Fix `llama-cpp` and `hf-inference-endpoints` extras in `pyproject.toml`

plaguss:
This PR solves a small problem when installing the following extras (there where some version/naming errors):

```sh
pip install -e ".[hf-inference-endpoints]"
pip install -e ".[llama-cpp]"
```


alvarobartt (approved):
Thanks a lot!

PR #140, opened 2023-12-04
Fix `llama_cpp_python` dependency check

plaguss:
There is a small naming problem with `llama_cpp` library that doesn't allow importing it. It should be named `llama_cpp_python`, otherwise the following evaluates to `False` even if the library is available:

```python
>>> from distilabel.utils.imports import _LLAMA_CPP_AVAILABLE
>>> _LLAMA_CPP_AVAILABLE
False
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ htmlcov/
 *.swp
 .DS_Store
 site/
+
+# Jupyter Notebook
+*.ipynb_checkpoints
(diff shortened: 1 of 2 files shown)


alvarobartt (approved):

alvarobartt:
LGTM! Just a nit, I think you're using an email to push the commits that's not identified and your commits are not linked to your account apparently, in case you want to fix it, make sure that your email is within `user.email` when `git config --list` (if not just `git config --global user.email agustin@argilla.io`)

image

PR #152, opened 2023-12-13
Adds `CritiqueTask`, `UltraCMTask` and more

### LINKED ISSUE #77: [FEATURE] Introduce Critique Task (UF, JudgeLM, Prometheus)
The idea is to introduce a new task to add LLM-As-Judge / Critique models that assess a single response individually instead of a list of generations together.

Ideally, this should work for:

- `len(generations)=1`
- `len(generations)>1`: needs an architectural change? The idea is to pass several generations and run the critique over each generation individually.

Target models:

- UF (used by Zephyr train_prefs)
- JudgeLM 
- Prometheus: high potential as it would mean getting rid of GPT4 as labeller

alvarobartt:
## Description

This PR adds the `CritiqueTask` as well as the subclass `UltraCMTask` which comes with the formatting expected by UltraCM 13B, which is a 13B LLM from OpenBMB, released as part of the UltraFeedback technical report, intended for critique / judge evaluation of completions for a given prompt with a pre-defined formatting.

Additionally, the `seed` arg has been included within `LlamaCppLLM`, as `seed=-1` has the same effect as `do_sample` in other LLMs such as the 🤗`transformers` ones.

Finally, the GGUF quantization of https://huggingface.co/openbmb/UltraCM-13b is available at https://huggingface.co/alvarobartt/UltraCM-13B-GGUF, since otherwise the model cannot be used locally on consumer hardware (thanks again `llama.cpp` <3)

Closes #77

## Disclaimer

The prompt has been modified as follows (original has been adapted to Jinja2 formatting):

```diff
Given my answer to an instruction, your role is to provide specific and constructive feedback for me. You should find the best way for me to learn from your feedback and improve my performance. 

You should consider multiple aspects of my answer, including helpfulness, truthfulness, honesty, and to what extent the answer follows instructions.
---

### Instruction
{{ instruction }}

### Answer
{{ completion }}
---

Please act as a teacher and provide specific and constructive feedback. Besides describing the weaknesses of the answer, you should also provide specific suggestions to guide me toward understanding how to improve. Please note, however, that your suggestions should help me better complete the instructions, but you should not introduce new requirements that are not mentioned in the instructions. Your feedback should focus on enhancing my ability to think critically and respond accurately. However, never explicitly provide the reference answer, nor do polite phrases be required. Only respond with concise feedback in chat style. Finally, score the overall quality of the answer from 1 to 10, where 1 is the worst and 10 is the best.

*Format*
### Feedback
Overall Score: [1-10]
[Your feedback]

---

- ### Feedback
- Overall Score: 
```

The modification has been done as otherwise the model was not performing correctly in most of the cases, and the response was poorly formatted, so tweaking a bit the prompt leads to better results, inducing a bit the start of the assistant response.

### DIFF
--- a/src/distilabel/tasks/critique/ultracm.py
+++ b/src/distilabel/tasks/critique/ultracm.py
@@ -1,0 +1,55 @@
+# Copyright 2023-present, Argilla, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from typing import ClassVar
+
+from typing_extensions import TypedDict
+
+from distilabel.tasks.base import get_template
+from distilabel.tasks.critique.base import CritiqueTask
+
+_ULTRACM_TEMPLATE = get_template("ultracm.jinja2")
+
+
+class UltraCMOutput(TypedDict):
+    """A `TypedDict` matching the output format of UltraCM."""
+
+    score: float
+    critique: str
+
+
+@dataclass
+class UltraCMTask(CritiqueTask):
+    __jinja2_template__: ClassVar[str] = _ULTRACM_TEMPLATE
+
+    system_prompt: str = (
+        "User: A one-turn chat between a curious user and an artificial intelligence"
+        " assistant. The assistant gives helpful, very detailed, and polite answers to"
+        " the user's questions.</s>"
+    )
+
+    def generate_prompt(self, instruction: str, completion: str) -> str:
+        render_kwargs = {
+            "instruction": instruction,
+            "completion": completion,
+        }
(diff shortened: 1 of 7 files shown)


src/distilabel/tasks/critique/ultracm.py
```diff
@@ -0,0 +1,59 @@
...
+        " assistant. The assistant gives helpful, very detailed, and polite answers to"
+        " the user's questions.</s>"
+    )
+
+    def generate_prompt(self, instruction: str, completion: str) -> str:
+        render_kwargs = {
+            "instruction": instruction,
+            "completion": completion,
+        }
+        return f"{self.system_prompt}\nUser: {self.template.render(**render_kwargs)}</s>\nAssistant: ### Feedback\nOverall Score: "
```
gabrielmbmb, line 49: We should return a `Prompt` here, right? We can add an `if` condition in the `Prompt.format_as` for `UltraCM`. I guess this class is meant to be used only with the `UltraCM` model, but I would prefer to not add this specificity and follow the `Task` interface.

gabrielmbmb (changes requested):
Hi, @alvarobartt! It LGTM, just a comment about the `generate_prompt` method of the `UltraCMTask`.

src/distilabel/tasks/critique/ultracm.py
alvarobartt: Indeed for e.g. Llama2 we're doing the same, as that's constrained to a model, and the UltraCM format is not a format per se, so I would keep it this way for the moment, as the `LLM` is ready to support both `str` and `Prompt`, and returning a `Prompt` here just adds complexity for something that's model specific and won't be used in any other scenario outside this one :/

gabrielmbmb (approved):

PR #158, opened 2023-12-16
Fix `_build_dataset` as `processed_labels` were ignored

plaguss:
## Description

We renamed a variable to `processed_labels` as part of #151 which was causing an error in the `Pipeline._build_dataset`, this PR solves the problem.

### DIFF
--- a/src/distilabel/pipeline.py
+++ b/src/distilabel/pipeline.py
@@ -522,15 +522,7 @@ class Pipeline:
         _dataset = Dataset(
             arrow_table=dataset.flatten_indices().data, split=Split.TRAIN
         )
-        # If no labeller is provided, the labels will be empty, so we create a fake dict
-        if not labels:
-            poped_label = {}
-        else:
-            poped_label = labels.pop(0)
-            # When we have futures, we cannot apply the operator **
-            if isinstance(poped_label, Future):
-                poped_label = poped_label.result()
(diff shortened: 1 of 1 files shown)


src/distilabel/pipeline.py
```diff
@@ -522,7 +522,15 @@ def _build_dataset(  # noqa: C901
...
-        _dataset = _dataset.map(lambda _: {**generations.pop(0), **labels.pop(0)})  # type: ignore
+        # If no labeller is provided, the labels will be empty, so we create a fake dict
+        if not labels:
+            poped_label = {}
+        else:
+            poped_label = labels.pop(0)
+            # When we have futures, we cannot apply the operator **
+            if isinstance(poped_label, Future):
+                poped_label = poped_label.result()
+        _dataset = _dataset.map(lambda _: {**generations.pop(0), **poped_label})  # type: ignore
```
alvarobartt, line 533: Isn't the issue that in https://github.com/argilla-io/distilabel/pull/151 the `labels` variable has been renamed to `processed_labels`? IMO the following should do the work:

```suggestion
        _dataset = _dataset.map(lambda _: {**generations.pop(0), **processed_labels.pop(0)}) # type: ignore
```

plaguss: you were right, didn't noticed that!

alvarobartt (approved):

PR #159, opened 2023-12-18
Add `to_argilla_{dataset,record}` methods in `TextGenerationTask`

alvarobartt:
## Description

This PR adds the `to_argilla_dataset` and `to_argilla_record` methods to the `TextGenerationTask` so that the generations can also be exported to Argilla.

By default those will contain `RatingQuestion`s from 1-10 on the quality of the generations based on the given inputs, and as we're not labelling anything it won't come with pre-filled suggestions as the `PreferenceTask` did.

Additionally, the `SelfInstructTask` has been fixed as the Jinja2 template used was not working as expected, the spacing was a bit off and the `num_examples` variable was not used and replaced with `num_instructions`; introduced at https://github.com/argilla-io/distilabel/pull/109.


ignacioct (approved):
LGTM

PR #163, opened 2023-12-18
Add `test` workflow

gabrielmbmb:
## Description

Add a simple `test` workflow, that will check that the lint and format is correct and then it will run the tests. In addition, fix some issues with type hints impeding to use the package with `python<3.9`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "Jinja2 >= 3.1.2",
     "rich >= 13.5.0",
     "tenacity >= 8",
-    "importlib-resources >=6.1.1; python_version < '3.9'",
+    "importlib-resources >= 6.1.1; python_version < '3.9'",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "Jinja2 >= 3.1.2",
     "rich >= 13.5.0",
     "tenacity >= 8",
+    "importlib-resources >=6.1.1; python_version < '3.9'",
```
alvarobartt, line 28: formatting nit

```suggestion
    "importlib-resources >= 6.1.1; python_version < '3.9'",
```

src/distilabel/tasks/base.py
```diff
@@ -12,7 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import importlib.resources as importlib_resources
+import sys
+
+if sys.version_info < (3, 9):
+    import importlib_resources
+else:
+    import importlib.resources as importlib_resources
```
alvarobartt, line 20:
```suggestion
import sys

if sys.version_info < (3, 9):
    import importlib_resources
else:
    import importlib.resources as importlib_resources

```

alvarobartt (approved):
LGTM, just some minor formatting nits

PR #167, opened 2023-12-19
Randomise generations order

gabrielmbmb:
## Description

This PR updates the `Pipeline._process_batch_generations` method to randomise the order of the generations for a given input. This way, we avoid adding prompt positional bias for the labeller, that could lead to certain model getting better/worse scores by the labeller by the position of its text generation in the prompt.


alvarobartt (approved):
LGTM, but maybe we could add a flag to control this behaviour? Because it may be unintended in certain scenarios

PR #169, opened 2023-12-19
Add custom `to_argilla_{dataset,record}` to `SelfInstructTask`

alvarobartt:
## Description

This PR adds a custom formatting for both the `FeedbackDataset` and the `FeedbackRecord` via `to_argilla_dataset` and `to_argilla_record`, respectively, for the `SelfInstructTask`, as re-using the formatting from the `TextGenerationTask` was not intuitive when annotating it via Argilla.

This PR is the result of a follow up discussion with @ignacioct, @plaguss and @dvsrepo

### DIFF
--- a/src/distilabel/tasks/text_generation/self_instruct.py
+++ b/src/distilabel/tasks/text_generation/self_instruct.py
@@ -12,12 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
+import warnings
 from dataclasses import dataclass
-from typing import Dict, List
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from distilabel.tasks.base import get_template
 from distilabel.tasks.prompt import Prompt
 from distilabel.tasks.text_generation.base import TextGenerationTask
+from distilabel.utils.argilla import infer_fields_from_dataset_row
+from distilabel.utils.imports import _ARGILLA_AVAILABLE
+
+if _ARGILLA_AVAILABLE:
+    import argilla as rg
+
+if TYPE_CHECKING:
+    from argilla.client.feedback.dataset.local.dataset import FeedbackDataset
+    from argilla.client.feedback.schemas.records import FeedbackRecord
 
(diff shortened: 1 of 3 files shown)


src/distilabel/tasks/text_generation/self_instruct.py
```diff
@@ -79,6 +90,97 @@ def generate_prompt(self, input: str) -> Prompt:
...
+        return ["instructions"]
+
     def parse_output(self, output: str) -> Dict[str, List[str]]:
         """Parses the output of the model into the desired format."""
-        return {"generations": output.split("\n")}
+        pattern = re.compile(r"\d+\.\s+(.*?)\n")
+        print(pattern.findall(output))
+        return {"instructions": pattern.findall(output)}
+
+    def to_argilla_dataset(self, dataset_row: Dict[str, Any]) -> "FeedbackDataset":
```
gabrielmbmb, line 103: Should we check here if argilla is installed?

gabrielmbmb (approved):

src/distilabel/tasks/text_generation/self_instruct.py
alvarobartt: No need to, as we're checking it in the `to_argilla` method already, and these methods are not intended to be used outside the `to_argilla` method of `CustomDataset`

src/distilabel/tasks/text_generation/self_instruct.py
```diff
@@ -79,6 +90,97 @@ def generate_prompt(self, input: str) -> Prompt:
...
 
+    @property
+    def output_args_names(self) -> List[str]:
+        return ["instructions"]
+
     def parse_output(self, output: str) -> Dict[str, List[str]]:
         """Parses the output of the model into the desired format."""
-        return {"generations": output.split("\n")}
+        pattern = re.compile(r"\d+\.\s+(.*?)\n")
+        print(pattern.findall(output))
```
alvarobartt, line 100:
```suggestion
```

PR #172, opened 2023-12-20
Fix `SelfInstructTask.{parse_output,to_argilla_record}` methods and `_build_dataset`

alvarobartt:
## Description

This PR adds some follow-up fixes for the `SelfInstructTask` as the `parse_output` method was strict with the formatting and it was expecting spaces to be right after each numbered list identifier, which was not the case in every output, so it has been loosened so that the space is optional. Additionally, this PR also adds some more consistency and robustness to the `to_argilla` method so that if the `to_argilla_record` method fails for any task, the error is handled and it continues until the loop over all the records ends; while also raising a `ValueError` if `SelfInstructTask.to_argilla_record` fails to generate the records.

Kudos to @ignacioct for spotting the issues related to `SelfInstructTask` and reporting right away!

Additionally, this PR also fixes an issue related to the `_build_dataset` method since the `map` apparently is not merging the labels within the dataset in order, the reason is not clear yet, but could be either due to the recently included `random` import or due to some `multiprocessing` happening in the background during the `map` method of `datasets`. So on, the current approach is more consistent.

### DIFF
--- a/src/distilabel/dataset.py
+++ b/src/distilabel/dataset.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
 from typing import TYPE_CHECKING, Union
 
 from datasets import Dataset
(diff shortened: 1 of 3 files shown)


ignacioct (approved):
Quick and right on the spot! Works on my side

PR #173, opened 2023-12-20
Fix `results` didn't have same order as `futures`

gabrielmbmb:
## Description

This PR fix a bug that was introduced in #164. The order of the `results` with respect of the `futures` was not being preserver in the `when_all_complete` function (kudos to @alvarobartt for spotting the issue).


alvarobartt (approved):
LGTM! I've already verified that it works fine, thanks for the quick fix

image

Issue #193, opened 2023-12-22
[FEATURE] Add checkpoint strategy for the dataset


plaguss:
**Is your feature request related to a problem? Please describe.**
I would like to save the dataset to disk during the generation process, it can be handy for long running pipelines to not loose the already generated data.

**Describe the solution you'd like**
A way to save every N generations the dataset to disk automatically.

**Describe alternatives you've considered**

**Additional context**

plaguss:
Closed with #194

PR #194, opened 2023-12-22
Feature dataset checkpoint strategy

### LINKED ISSUE #168: could there be a feature to save while generating？
It's great that this works, but could there be a feature to save while generating, so that the pipeline doesn't go wrong in the middle and the previous data runs in vain?

plaguss:
## Description

This PR modifies `enable_checkpoint` in the `Pipeline.generate` to allow a more behaviour during the generation (saving while generating).

The following example allows saving the `CustomDataset` every 100 generations to disk.
By default, it will use DatasetCheckpoint(save_frequency=-1), which replicates the previous behaviour of `enable_checkpoints=True`.

```python
from distilabel.utils.dataset import DatasetCheckpoint

pipe.generate(
    ...,
    dataset_strategy=DatasetCheckpoint(save_frequency=100)
)
```

Closes #168  and #193.


src/distilabel/dataset.py
```diff
@@ -13,10 +13,12 @@
 # limitations under the License.
 
 import warnings
+from os import PathLike
 from typing import TYPE_CHECKING, Union
```
gabrielmbmb, line 17:
```suggestion
from typing import TYPE_CHECKING, Any, Union
```

src/distilabel/dataset.py
```diff
@@ -90,3 +92,31 @@ def to_argilla(self) -> "FeedbackDataset":
                     stacklevel=2,
                 )
         return rg_dataset
+
+    def save_to_disk(self, dataset_path: PathLike, **kwargs) -> None:
```
gabrielmbmb, line 96:
```suggestion
    def save_to_disk(self, dataset_path: PathLike, **kwargs: Any) -> None:
```

src/distilabel/dataset.py
```diff
@@ -90,3 +92,31 @@ def to_argilla(self) -> "FeedbackDataset":
...
+        Args:
+            dataset_path: Path to the dataset.
+            **kwargs: Additional arguments to be passed to `datasets.Dataset.save_to_disk`.
+        """
+        super().save_to_disk(dataset_path, **kwargs)
+        if self.task is not None:
+            save_task_to_disk(dataset_path, self.task)
+
+    @classmethod
+    def load_from_disk(cls, dataset_path: PathLike, **kwargs):
```
gabrielmbmb, line 108:
```suggestion
    def load_from_disk(cls, dataset_path: PathLike, **kwargs: Any):
```

src/distilabel/utils/dataset.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def save_task_to_disk(path: Path, task: "Task") -> None:
+    """Saves a task to disk.
+
+    Args:
+        path: The path to the task.
+        task: The task.
+    """
+    task_path = path / "task.pkl"
```
gabrielmbmb, line 31: we can create a constant `TASK_FILE_NAME = "task.pkl"`

src/distilabel/utils/dataset.py
```diff
@@ -0,0 +1,87 @@
...
+        Args:
+            step (int): The number of records generated.
+
+        Returns:
+            bool: Whether a checkpoint should be done.
+        """
+        if self.save_frequency == -1:
+            return False
+
+        if (step - self._total_checks * self.save_frequency) // self.save_frequency:
```
gabrielmbmb, line 84: I think this should be enough:

```suggestion
        if step % self.save_frequency == 0:
```

src/distilabel/utils/dataset.py
```diff
@@ -0,0 +1,87 @@
...
+    task_path = path / "task.pkl"
+    if not task_path.exists():
+        raise FileNotFoundError(f"The task file does not exist: {task_path}")
+    with open(task_path, "rb") as f:
+        task = pickle.load(f)
+    return task
+
+
+@dataclass
+class DatasetCheckpoint:
```
gabrielmbmb, line 54: I think we could move this class to `src/distilabel/dataset.py` or at least make it importable like `from distilabel.dataset import DatasetCheckpoint`

src/distilabel/pipeline.py
```diff
@@ -623,20 +625,39 @@ def _generate(  # noqa: C901
...
+                            ds = self._build_dataset(
+                                dataset,
+                                generations=generations,
+                                labels=labels,
+                                batch_size=batch_size,
+                            )
+                            ds.save_to_disk(
+                                checkpoint_strategy.path,
+                                **checkpoint_strategy.extra_kwargs,
+                            )
```
gabrielmbmb, line 641: I think this block within the `if self.generator is not None` and condition and the one below within the `if self.labeller is not None` condition could be the same, right? We could just simply:

```python
for batch_i, rows in enumerate(dataset.iter(batch_size=batch_size), start=1):
    if self.generator is not None: ...
    if self.labeller is not None: ...
    if checkpoint_strategy and checkpoint_strategy.do_checkpoint(batch_i * batch_size): # do checkpoint
```

gabrielmbmb (changes requested):
Hi @plaguss! Much needed feature :) it looks good, I left some comments

src/distilabel/utils/dataset.py
plaguss: Yes, I would go with the first option

src/distilabel/pipeline.py
plaguss: yes, thanks!

src/distilabel/utils/dataset.py
plaguss: I've added additional tests for this one [here](https://github.com/argilla-io/distilabel/pull/194/files#diff-26199baaf391361625bd6773b9df4ed2677df4680a9650d2731510adec4898eaR47-R59), the last one is a weird option, but if someone uses some combinations of `batch_size` and `save_frequency` this approach would not save the dataset.

gabrielmbmb (approved):

plaguss:
Added a small reference in the docs, inside a new section of the `Pipeline` from the concept guides

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "tenacity >= 8",
     "importlib-resources >= 6.1.1; python_version < '3.9'",
     "multiprocess",
+    "dill >= 0.3.7",
```
alvarobartt, line 30: I'm not a big fan of `dill`, `pickle` and similar, so maybe we can explore other ways for saving/dumping the `Task` configuration?

src/distilabel/dataset.py
```diff
@@ -90,3 +94,74 @@ def to_argilla(self) -> "FeedbackDataset":
...
+        extra_kwargs (dict[str, Any]): Additional kwargs to be passed to the `save_to_disk` method of the Dataset.
+
+    Examples:
+        >>> from distilabel.dataset import DatasetCheckpoint
+        >>> # Save the dataset every 10% of the records generated.
+        >>> checkpoint = DatasetCheckpoint(save_frequency=len(dataset) // 10)
+        >>> # Afterwards, we can access the checkpoint checkpoint.path.
+    """
+
+    path: Path = Path.cwd() / "dataset_checkpoint"
```
alvarobartt, line 145: Maybe we can follow the standard and name this `ckpt/` instead?

src/distilabel/pipeline.py
```diff
@@ -527,12 +527,54 @@ def _generate(  # noqa: C901
...
+            ...     tokenizer="meta-llama/Llama-2-7b-chat-hf",
+            ...     task=Llama2TextGenerationTask(),
+            ... )
+            >>> labeller = OpenAILLM(
+            ...     model="gpt-3.5-turbo",
+            ...     task=UltraFeedbackTask.for_text_quality(),
+            ... )
+            >>> pipeline = Pipeline(generator=generator, labeller=labeller)
+            >>> dataset = pipeline.generate(dataset=..., num_generations=1, batch_size=1)
+        """
```
alvarobartt, line 577: This was removed to avoid having duplicated docstrings, as `_generate` won't be used by the users, but `generate` instead, so we'll only maintain the `generate` docstrings

```suggestion
        `Pipeline`."""
```

alvarobartt (commented):
LGTM even though I'm still pending to test it myself!

Also a general comment is that I'd like to discuss on the naming with the rest of the team, as IMO we could just have `checkpoints_path` and `checkpoints_freq` or something similar, rather than asking the user to instantiate a class, while we may use/re-use the class internally, but to avoid the user for instantiating multiple classes within the same function.

src/distilabel/dataset.py
plaguss: Sure!

pyproject.toml
plaguss: As discussed internally, will review this in the future😄

docs/technical-reference/pipeline.md
```diff
@@ -116,6 +116,28 @@ Then, we will load the dataset and call the `generate` method of the pipeline. F
 
 With a few lines of code, we have easily generated a dataset with 2 generations per input, using 4 different `LLM`s, and labelled the generations using GPT-4. You can check the full code [here](https://github.com/argilla-io/distilabel/blob/main/examples/pipeline-preference-dataset-llmpool.py).
 
+### Dataset checkpoints
+
+With long pipelines, it may be useful to review the dataset during the process, just to keep track and stop the process, or have it saved in case something fails before obtaining the final dataset. We can use the `checkpoint_strategy` in `Pipeline.generate` method for this end:
```
ignacioct, line 121: I don't understand the sentence "just to keep track and stop the process".

src/distilabel/pipeline.py
```diff
@@ -425,13 +425,13 @@ def _build_dataset(  # noqa: C901
...
-                        for _ in range(len(dataset) - len(generations))
-                    ]
-                )
+            # if len(generations) < len(dataset):
+            #     generations.extend(
+            #         [
+            #             {key: None for key in generator_column_names}
+            #             for _ in range(len(dataset) - len(generations))
+            #         ]
+            #     )
```
ignacioct, line 434: Do we need this data commented?

src/distilabel/pipeline.py
```diff
@@ -611,26 +612,51 @@ def _generate(  # noqa: C901
...
                         dataset,
                         generations=generations,
                         labels=labels,
                         batch_size=batch_size,
                     )
 
+            if checkpoint_strategy and checkpoint_strategy.do_checkpoint(
+                batch_i * batch_size
+            ):
+                logger.info(f"Saving dataset up to batch {batch_i}...")
```
ignacioct, line 635: Would it useful for the user to print the writing file here?

src/distilabel/pipeline.py
```diff
@@ -611,26 +612,51 @@ def _generate(  # noqa: C901
...
+        ds = self._build_dataset(
             dataset, generations=generations, labels=labels, batch_size=batch_size
         )
+        if checkpoint_strategy:
+            logger.info("Saving final dataset...")
+            ds.save_to_disk(
+                checkpoint_strategy.path,
+                **checkpoint_strategy.extra_kwargs,
+            )
+
```
ignacioct, line 658:
```suggestion
        if checkpoint_strategy:
            ds.save_to_disk(
                checkpoint_strategy.path,
                **checkpoint_strategy.extra_kwargs,
            )
            logger.info("Final dataset saved.")

```

ignacioct (commented):
Just a couple of personal doubts and minor suggestions, but looks really nice.

docs/technical-reference/pipeline.md
plaguss: It should be an or. But anyway I see that it doesn't make much sense. Removing that altogether

src/distilabel/pipeline.py
plaguss: I forgot about that. I'm not sure of the use of that lines, but the code fails with that code. If that code is activated the dataset isn't correctly generated, it would only save the dataset up to a given batch, loosing that part of the dataset already generated. Commenting that part, everything seems to work fine.
@alvarobartt, @gabrielmbmb, could you help me here to know what this piece of code is for?

src/distilabel/pipeline.py
plaguss: do you mean the path to the file? Will do it!

src/distilabel/pipeline.py
plaguss: Will update this one with the checkpoint path as suggested in the previous review 😄

src/distilabel/pipeline.py
gabrielmbmb: This is for a very unlikely (I think it's not possible anymore for it to happen) case in which we haven't generated texts for all the rows, but we can remove this block because the solution is not good either.

plaguss: In that case erased as it never happened

PR #195, opened 2023-12-22
Fix `PrometheusTask` and `UltraCMTask` could not be chained with `TextGenerationTask`

gabrielmbmb:
## Description

The arguments of `generate_prompt` method for both `PrometheusTask` and `UltraCMTask` classes were not aligned with the ones generated by the `TextGenerationTask`, not allowing to have a `Pipeline` with a `generator` using the `TextGenerationTask` and a `labeller` using either `PrometheusTask` or `UltraCMTask`.

In addition, all the `generate_prompt` methods have been updated to accept extra keyword arguments and ignore them (`**_`). This is useful and convenient for the cases in which a `Dataset` contains more columns than the required by the task. Before if the input dataset contained additional columns not used by the task it would cause an error.

### DIFF
--- a/examples/pipeline-prometheus.py
+++ b/examples/pipeline-prometheus.py
@@ -50,8 +50,8 @@ if __name__ == "__main__":
 
     dataset = Dataset.from_dict(
         {
-            "instruction": ["What's the capital of Spain?"],
-            "completion": ["Paris"],
+            "input": ["What's the capital of Spain?"],
+            "generations": ["Paris"],
             "ref_completion": ["Madrid"],
         }
     )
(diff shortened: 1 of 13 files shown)


dvsrepo:
Thanks a lot @gabrielmbmb !

I also noticed yesterday the UltraCM didn't return a Prompt like other tasks but a str. And also the Prometheus one had a hardcoded prompt format as llama, this would also cause issues integrating with the rest of the stack. I guess the reason is these OSS models are very sensitive to prompt templates but we need to discuss how to make this more interoperable or at least inform the user.

In general this wouldn't have been overlooked, if we have added a meaningful example showing the full workflow and likewise adding a very brief doc piece.

PR #197, opened 2023-12-22
Add draft with social cards

### LINKED ISSUE #183: [DOCS] Add social card plugin
We want to generate nice social cards for sharing in social media. 

We should use this plugin:

https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/#built-in-social-plugin

I'd suggest something minimal, nice contrast between background, fonts, and logo. 

Make sure the font looks nice and big and we show the ⚗️ logo we use in svg for the docs. 

It should be set up so every page shows a custom title and description, for example:

https://distilabel.argilla.io/latest/technical-reference/llms/

This page should show LLMs and a description. If it doesn't happen automatically we might need to setup metadata on each page but I think it should happen automatically.

There's a video explaining how to setup this and debug it using the cached generated images, @plaguss feel free to send here some initial alternatives so we can choose.

plaguss:
## Description

I push this PR so we can take a look at the default social cards. I still need to discover how to add a custom description to all the cards. For the docs we have created in `/docs` folder we can write it directly in the .md file, but still need to know how to do it for the reference we have from the docstrings.

Would close #183


plaguss:
I have added a sentence/small description for the main pages in our docs, for example:
image
The remaining pages will render with the default content:
image
Let me know what you think of those and we can iterate.

dvsrepo (approved):
Awesome!

PR #199, opened 2023-12-27
Update `generate_prompt` in `Task` subclasses to always return `Prompt`

### LINKED ISSUE #186: [BUG] UltraCMTask fails when using it with OpenAILLM
I would expect the UltraCMTask to work with OpenAILLM or at least to throw an error to tell me that this is not supported.

Unfortunately, it fails with the following:

```python
from datasets import Dataset
from distilabel.llm import OpenAILLM
from distilabel.tasks.critique.ultracm import UltraCMTask


pipeline = Pipeline(
        labeller=OpenAILLM(
            task=UltraCMTask(),
            model="gpt-4",
            openai_api_key=userdata.get('OPENAI_API_KEY'),
            temperature=1.0,
        ),
    )

dataset = Dataset.from_dict(
        {
            "instruction": ["What's the capital of Spain?"],
            "completion": ["Madrid"],
        }
    )

dataset = pipeline.generate(
        dataset,  # type: ignore
        display_progress_bar=True,
)

#error  
*Format*
### Feedback
Overall Score: [1-10]
[Your feedback]

---
</s>
Assistant: ### Feedback
Overall Score: ` is of `type=<class 'str'>`, but it must be of `type=<class 'list'>`, so make sure that `task.generate_prompt` returns a `<class 'list'>` or that the `formatting_fn` formats the prompt as a  `<class 'list'>`.

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
RuntimeError: `Pipeline.generate` failed during labelling step. Setting `enable_checkpoints=True` is recommended!

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
[/usr/local/lib/python3.10/dist-packages/distilabel/pipeline.py](https://localhost:8080/#) in dry_run(self, dataset)
    703         except Exception as e:
    704             self._teardown()
--> 705             raise RuntimeError(
    706                 f"`Pipeline.generate` failed during the dry run over {dataset[0]} with exception: {e}"
    707             ) from e

RuntimeError: `Pipeline.generate` failed during the dry run over {'instruction': "What's the capital of Spain?", 'completion': 'Madrid'} with exception: `Pipeline.generate` failed during labelling step. Setting `enable_checkpoints=True` is recommended!
```

I would assume our tasks should support swapping LLMs, if this is not feasible we need to rethink the interaction between Task and LLM. From what I see other tasks return a Prompt instead of string, and this is a relatively easy to follow prompt and output for OpenAI models, which will allow avoiding the usage of a heavy OSS/local model.

alvarobartt:
## Description

This PR ensures that all the methods `generate_prompt` defined within the abstract class `Task`, return a `Prompt` with no formatting, as the formatting is handled internally by the `Prompt` via the `format_as` method, which is used by the user via the arg `prompt_format` within each `LLM`. This way, we ensure that the `Task` implementations are LLM-agnostic, meaning that any task format can be used with any LLM.

So on, the above implies that both `OpenAITextGenerationTask` and `Llama2TextGenerationTask` have been removed too, as those were constrained to the formatting of an LLM, in particular OpenAI and Llama2, respectively.

Closes #186


examples/pipeline-preference-dataset-llmpool.py
```diff
@@ -108,7 +108,8 @@ def load_gpt_4(task: UltraFeedbackTask) -> LLM:
             ]
         ),
         labeller=ProcessLLM(
-            task=UltraFeedbackTask.for_instruction_following(), load_llm_fn=load_gpt_4
+            task=UltraFeedbackTask.for_instruction_following(),
+            load_llm_fn=load_gpt_4,  # type: ignore
```
gabrielmbmb, line 112: instead of adding this `# type: ignore`, could you change the `task` argument type of `load_gpt_4` function to `Task`?

gabrielmbmb (approved):

PR #203, opened 2023-12-29
Add `ChatTask` for multi-turn and follow-up generation

### LINKED ISSUE #157: [FEATURE] Introduce FollowUp generator
## Description
A high impact task for distilabel is one that generates follow up turns or multi-turn dialogues (which then can be criticized/ranked 

Given a conversation (or at least a prompt+response pair), the generator will generate a follow up message (from the user role). 

Ideally the input would be a standard conversation/list of messages format (like the one we use in uf, zephyr, etc.). This format can be used to build the generator prompt and ask it to generate a follow up message.

This can be developed in parallel or before #130 

## Open questions:

1. Should we also build/test a generator that generates not only the follow up message but also the response? My concern is that this is less composable than `Instructions` -> `MultiTurner` -> and then use our current pipelines: `Response generator` -> Labeler. 
2. If we do the above, can we define the number of turns to generate? My concern is to ask too much to this component and end up with Alpaca-level quality (bad) turns.
3. Is this a good image for the `MultiTurner` task?

multiturner

alvarobartt:
## Description

This PR adds the `ChatTask` to easily define and prepare a task to generate follow up messages or even to use the multi-turn information to generate a response based on the previous context.

The `ChatTask` produces a `Chat` following the pre-defined formatting, but this should be discussed before merging, as an alignment between `Prompt` and `Chat` is much needed, as well as a proper identification of the different purposes to serve i.e. `Chat` and `Instruct`, and what's the best approach to cover those while still providing some formatting functions.

Closes #157

## Example

```python
import os
from uuid import uuid4

import argilla as rg
from datasets import Dataset
from distilabel.tasks.chat import ChatTask

if __name__ == "__main__":
    dataset = Dataset.from_dict(
        {
            "messages": [
                [
                    {
                        "role": "system",
                        "content": "You are a helpful assistant that always answers like a pirate.",
                    },
                    {
                        "role": "user",
                        "content": "What's the most terrifying creature in the sea?",
                    },
                ]
            ],
        }
    )

    generator = OpenAILLM(
        model="gpt-4",
        task=ChatTask(input_format="openai", output_format="openai"),
        openai_api_key=os.getenv("OPENAI_API_KEY"),
        max_new_tokens=1024,
        prompt_format=None,
    )

    pipeline = Pipeline(generator=generator)
    dataset = pipeline.generate(
        dataset=dataset,
        shuffle_before_labelling=False,
        num_generations=2,
        skip_dry_run=True,
        display_progress_bar=True,
    )  # type: ignore

    rg_dataset = dataset.to_argilla()

    rg.init(
        api_url=os.getenv("ARGILLA_API_URL"),
        api_key=os.getenv("ARGILLA_API_KEY"),
    )

    rg_dataset.push_to_argilla(name=f"chat-{uuid4()}", workspace="admin")

```

### DIFF
--- a/examples/pipeline-chat.py
+++ b/examples/pipeline-chat.py
@@ -1,0 +1,66 @@
+# Copyright 2023-present, Argilla, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from uuid import uuid4
+
+import argilla as rg
+from datasets import Dataset
+from distilabel.llm.openai import OpenAILLM
+from distilabel.pipeline import Pipeline
+from distilabel.tasks.chat import ChatTask
+
+if __name__ == "__main__":
+    dataset = Dataset.from_dict(
+        {
+            "messages": [
+                [
+                    {
+                        "role": "system",
+                        "content": "You are a helpful assistant that always answers like a pirate.",
+                    },
+                    {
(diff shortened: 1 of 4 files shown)


davidberenstein1957:
Ideally, we want to generate data with follow-up questions.

PR #204, opened 2024-01-02
Add `VertexAILLM` & `VertexAIEndpointLLM` classes

gabrielmbmb:
## Description

This PR adds or modifies:

- `VertexAILLM` class, which allows to use an LLM from the Gemini API or Text/Code Generation API
- `VertexAIEndpointLLM` class, which allows to use a Vertex AI Endpoint (users can deploy Llama2, Mistral, etc or any model on this endpoints)
- Remove the `expected_output_type ` from `LLM._generate_prompts` method

### DIFF
--- a/src/distilabel/llm/google/vertexai.py
+++ b/src/distilabel/llm/google/vertexai.py
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+
+from tenacity import (
+    after_log,
+    before_sleep_log,
+    retry,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_random_exponential,
+)
 
 from distilabel.llm import LLM
 from distilabel.llm.utils import LLMOutput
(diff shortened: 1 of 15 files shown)


src/distilabel/llm/google/vertexai.py
```diff
@@ -0,0 +1,517 @@
...
+                objects.
+        """
+        contents = []
+        for input in inputs:
+            prompt = self.task.generate_prompt(**input)
+            contents.append(
+                [
+                    {"role": "user", "parts": [{"text": prompt.system_prompt}]},
+                    {"role": "model", "parts": [{"text": "Understood."}]},
+                    {"role": "user", "parts": [{"text": prompt.formatted_prompt}]},
```
alvarobartt, line 207: I'm a bit confused here, is the `model` message really required? Isn't it better to just concatenate the `system_prompt` and the `formatted_prompt` using a line-break instead as the only `user` message?

alvarobartt (approved):
Great PR, thanks for the effort! 💪🏻

One question that comes to my mind after seeing the struggle on the formatting is, isn't it better to move the formatting out of the `Prompt` class and assume the formatting is always going to be chat-like following OpenAI format, while still providing some helper functions to be provided as `prompt_formatting_fn` to e.g. go from plain text to Zephyr-like?

src/distilabel/llm/google/vertexai.py
gabrielmbmb: Yes, that works too.

PR #209, opened 2024-01-04
Add `ReplicateLLM`

### LINKED ISSUE #122: [FEATURE] Add support for Replicate endpoints
The idea is to implement a new LLM for using Replicate endpoints

Ideally it should cover `public` and private `deployments`
Draft here:

https://github.com/argilla-io/distilabel/pull/47

See an example of a HTTP-based implementation:

https://github.com/argilla-io/distilabel/blob/main/src/distilabel/llm/huggingface/inference_endpoints.py

alvarobartt:
## Description

This PR adds the `ReplicateLLM` class to easily use Replicate Endpoints within `distilabel`.

Closes #122 (as a follow up of #47)

## Things to discuss before merging

* Generation kwargs definition is poor, because it's constrained to the `cog` container used, and it's most of the times different per each model hosted in Replicate, which makes it harder to use via kwargs
* The fact that the pre/post processing happens internally in the `cog` container, the formatting of the prompt is tricky as sometimes that's expected to come pre-formatted on the client side, while other times that's handled by the container, but as that's controlled with kwargs, it's impossible to tell
* After some testing I think that the latency is better in other competitors, so maybe we can keep this PR as a draft until we figure out whether there's a clean way we can make it work fine with `distilabel` because as of now, it's really messy

### DIFF
--- a/examples/pipeline-replicate.py
+++ b/examples/pipeline-replicate.py
@@ -1,0 +1,54 @@
+# Copyright 2023-present, Argilla, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


dvsrepo:
Thanks @alvarobartt , great work and analysis. I agree we can leave it as draft pr for now and see if there's interest from users

PR #213, opened 2024-01-05
Add Anyscale endpoints

### LINKED ISSUE #121: [FEATURE] Add support for AnyScale inference endpoints

plaguss:
## Description

This PR adds [anyscale](https://docs.endpoints.anyscale.com/) endpoints on top of the OpenAI integration.

Would close #121


examples/anyscale-endpoints.py
alvarobartt, line 1: Maybe we can just align this example with the rest of the examples to also include the Argilla integration, and just remove the output of the generation?

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,122 @@
...
+    from distilabel.tasks.prompt import SupportedFormats
+
+logger = get_logger()
+
+
+class AnyscaleLLM(OpenAILLM):
+    def __init__(
+        self,
+        task: "Task",
+        model: str = "HuggingFaceH4/zephyr-7b-beta",
```
alvarobartt, line 36: Maybe the `model` shouldn't have a default value?

alvarobartt (approved):
LGTM, but shouldn't we add a check to ensure that `OpenAILLM` and `AnyscaleLLM` are not used within the same `Pipeline`? As AFAIK that would fail since the `OPENAI_BASE_URL` is modified as well as the `OPENAI_API_KEY`, right?

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,122 @@
...
+        prompt_formatting_fn: Union[Callable[..., str], None] = None,
+    ) -> None:
+        """Initializes the AnyscaleLLM class.
+
+        Args:
+            task (Task): the task to be performed by the LLM.
+            model (str, optional): the model to be used for generation. Defaults to "HuggingFaceH4/zephyr-7b-beta".
+            client (Union[OpenAI, None], optional): an OpenAI client to be used for generation.
+                If `None`, a new client will be created. Defaults to `None`.
+            openai_api_key (Union[str, None], optional): the OpenAI API key to be used for generation.
```
alvarobartt, line 55:
```suggestion
            api_key (Union[str, None], optional): the Anyscale API key to be used for generation.
```

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,122 @@
...
+            ...    'prompt_used': [{'role': 'system',
+            ...        'content': "You...
+        """
+        LLM.__init__(
+            self,
+            task=task,
+            num_threads=num_threads,
+            prompt_format=prompt_format,
+            prompt_formatting_fn=prompt_formatting_fn,
+        )
```
alvarobartt, line 99:
```suggestion
        super().__init__(
            task=task,
            num_threads=num_threads,
            prompt_format=prompt_format,
            prompt_formatting_fn=prompt_formatting_fn,
        )
```

alvarobartt:
> LGTM, but shouldn't we add a check to ensure that `OpenAILLM` and `AnyscaleLLM` are not used within the same `Pipeline`? As AFAIK that would fail since the `OPENAI_BASE_URL` is modified as well as the `OPENAI_API_KEY`, right?

I guess it would work fine as long as we don't rely on the environment variables, right? But unsure on how that's internally handled by OpenAI, will the provided key via arg override whatever is on the environment variable?

gabrielmbmb (changes requested):
LGTM! Could we add some unit tests to the `OpenAILLM` class similar to the ones from `VertexAILLM`?

plaguss:
> > LGTM, but shouldn't we add a check to ensure that `OpenAILLM` and `AnyscaleLLM` are not used within the same `Pipeline`? As AFAIK that would fail since the `OPENAI_BASE_URL` is modified as well as the `OPENAI_API_KEY`, right?
> 
> I guess it would work fine as long as we don't rely on the environment variables, right? But unsure on how that's internally handled by OpenAI, will the provided key via arg override whatever is on the environment variable?

Will review the handling from the OpenAI side, from the Anyscale examples they assume we are just replacing one with the other, but didn't think about the case of the `LLMPool`.

tests/llm/test_anyscale.py
```diff
@@ -0,0 +1,96 @@
...
+
+@mock.patch("distilabel.llm.anyscale.OpenAI")
+@mock.patch(
+    "distilabel.llm.anyscale.OpenAILLM.available_models",
+    ["HuggingFaceH4/zephyr-7b-beta"],
+)
+class TestAnyscaleLLM:
+    def test_llm_anyscale(self, mock_anyscale):
+        llm = AnyscaleLLM(
+            model="HuggingFaceH4/zephyr-7b-beta",
```
alvarobartt, line 31: Is there any chance we use `argilla/notus-7b-v1` instead?

plaguss: In the tests there should be no problem, will update it this afternoon. But it's not available to run (I requested both notus and notux, but haven't received an answer yet). This are the supported ones https://docs.endpoints.anyscale.com/category/supported-models.

plaguss: > In the tests there should be no problem, will update it this afternoon. But it's not available to run (I requested both notus and notux, but haven't received an answer yet). These are the supported ones https://docs.endpoints.anyscale.com/category/supported-models.

Updated

PR #215, opened 2024-01-05
Add `TogetherInferenceLLM`

### LINKED ISSUE #95: [FEATURE] Add support for Together Inference endpoints
The idea would be to implement a new LLM to support inference with Together. 

See an example of a HTTP-based implementation:

https://github.com/argilla-io/distilabel/blob/main/src/distilabel/llm/huggingface/inference_endpoints.py

Ref: https://www.together.ai/blog/together-inference-engine-v1

alvarobartt:
## Description

This PR adds the `TogetherInferenceLLM` to support Together Inference Endpoints via their Python client. Also one example has been included under `examples/` and the documentation has been updated accordingly.

Additionally, this PR also fixes a bug within the Argilla integration developed to unwrap the rows when using the `LLMPool` that was affecting the `TextGenerationTask`, which is now solved.

Closes #95

### DIFF
--- a/src/distilabel/llm/together.py
+++ b/src/distilabel/llm/together.py
@@ -23,7 +23,6 @@ from distilabel.utils.imports import _TOGETHER_AVAILABLE
 
 if _TOGETHER_AVAILABLE:
     import together
-    from together.types import TogetherResponse
 
 if TYPE_CHECKING:
     from distilabel.tasks.base import Task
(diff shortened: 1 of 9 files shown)


src/distilabel/llm/together.py
```diff
@@ -0,0 +1,215 @@
...
+        logprobs: int = 0,
+        num_threads: Union[int, None] = None,
+        prompt_format: Union["SupportedFormats", None] = None,
+        prompt_formatting_fn: Union[Callable[..., str], None] = None,
+    ) -> None:
+        """Initializes the OpenAILLM class.
+
+        Args:
+            task (Task): the task to be performed by the LLM.
+            model (str, optional): the model to be used for generation. Defaults to "gpt-3.5-turbo".
```
plaguss, line 56:
```suggestion
            model (str, optional): the model to be used for generation.
```

plaguss (changes requested):
Awesome!

PR #221, opened 2024-01-08
Update `LICENSE` and add `LICENSE_HEADER`

davidberenstein1957:
We got this comment on X https://twitter.com/ImageDeeply/status/1743982093707710835
I copied the license file from https://github.com/argilla-io/argilla/edit/develop/LICENSE, which seems to work properly in the Argilla repo.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [2021] [Recognai S.L.]
+   Copyright [2023-present] [Argilla, Inc.]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -1,13 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [2021] [Recognai S.L.]
```
alvarobartt, line 189:
```suggestion
   Copyright [2023-present] [Argilla, Inc.]
```

alvarobartt (changes requested):
Fair, thanks! This would be a breaking change, since we inject the `LICENSE` on top of each file, could you create a `LICENSE_HEADER` with the previous stuff, and then change the path within `.pre-commit-config.yaml` so that's injected instead of `LICENSE`?

alvarobartt (approved):

PR #226, opened 2024-01-09
Add badges to readme

plaguss:
## Description

This PR includes two badges to include to share in a model card and show a dataset was built using `distilabel`.

Thanks @leiyre for the design ❤️


ignacioct:
Love 'em !

Issue #233, opened 2024-01-10
[FEATURE] Clickable anchors on mkdocs


ignacioct:
As @dvsrepo pointed out, we should add clickable anchors, like in Argilla's docs, so we can click a subsection and copy the URL

![Screenshot_20240110-155422](https://github.com/argilla-io/distilabel/assets/56955040/8d4436dd-4847-4ef3-93df-efc466b7b727)

PR #235, opened 2024-01-11
Docs/binders anchors

### LINKED ISSUE #234: [DOCS] Fix the binders in the tutorials and show them in the upper right corner, like in Argilla

Right now, we have to problems with the binders that need solving:

- [ ] Place binders automatically in the upper right corner, outside of the document, like we do in Argilla.
- [ ] Binder's link doesn't directly lead to the collab, but to a prior place where you have to download/choose collab.

ignacioct:
Closes #234 and #233 

As suggested by @dvsrepo , we have included clickable anchors to easily navigate and share URLs. I've also fixed the Open In Colab button using https://openincolab.com/ to generate the mirror URLs, now it opens directly the tutorial. 

I've been also looking into putting the Open In Colab button in the top-right corner, like in Argilla, but I think that is automatically generated by readthedocs, and I've been able to find any option similar in mkdocs. If somebody knows how to do it, I'll be super happy to implement it. 

Once this PR is pushed, I'll merge main into 0.3.0 to make the tutorial available, as suggested by @davidberenstein1957.

Thanks again to @plaguss for his help fighting the Conda 🐍

### DIFF
--- a/docs/tutorials/pipeline-notus-instructions-preferences-legal.ipynb
+++ b/docs/tutorials/pipeline-notus-instructions-preferences-legal.ipynb
@@ -6,6 +6,9 @@
    "source": [
     "# 🤗 Use Notus on inference endpoints to create a legal preference dataset\n",
     "\n",
+    "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/argilla-io/distilabel/blob/main/docs/tutorials/pipeline-notus-instructions-preferences-legal.ipynb) [![Open Source in Github](https://img.shields.io/badge/github-view%20source-black.svg)](https://github.com/argilla-io/distilabel/blob/main/docs/tutorials/pipeline-notus-instructions-preferences-legal.ipynb)\n",
+    "\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


davidberenstein1957:
Hi @ignacioct, good work on the anchors and button :) 

We've set that up automatically here for templating using nbsphinx docs: https://github.com/argilla-io/argilla/blob/0b91c8ff010d17e269c1969fdfe882449865af94/docs/_source/conf.py#L217

I assume it is possible with MkDocs too but the information seems lacking so don't spend too much time trying to reproduce it if you don't have a clear idea. https://www.mkdocs.org/user-guide/customizing-your-theme/#overriding-template-blocks

Also, I think we can add a open source in github button as within the added Argilla. 

I would add the button a bit higher in the tutorial. Maybe directly underneath the header?

Lastly, I would change the order of the tabs and place "tutorials" before "technical references". 
image

ignacioct:
@davidberenstein1957 everything that you mentioned is changed now! Everything I've tried to add the buttons on top hasn't worked, are we planning on staying with mkdocs? To add that issue to the backlog and maybe tackle it later

davidberenstein1957:
@ignacioct, I assume will so you can add it to the backlog as an issue.

ignacioct:
> @ignacioct, I assume will so you can add it to the backlog as an issue.

Then I'm merging now! Thanks :)

PR #237, opened 2024-01-11
Add Evol instruct task

### LINKED ISSUE #229: [FEATURE] add support for the `EvolInstructTask` from the `WizardLM` paper
The [WizardLM paper](https://arxiv.org/abs/2304.12244) recently introduced the `EvolInstructTask` and it would nice to have this integrated within distilabel too as proposed by @gabrielmbmb.

An example can be found [here](https://github.com/lcw99/evolve-instruct/blob/main/evolve.py)
Paper can be found [here](https://arxiv.org/abs/2304.12244)

N.A.

**Additional context**
This allows us to reproduce [deita](https://arxiv.org/abs/2312.15685)

plaguss:
## Description

This PR adds a new text generation task: *Evol-Instruct*.

Closes #229


src/distilabel/tasks/_templates/evol-instruct.jinja2
```diff
@@ -0,0 +1,33 @@
+{%- if evol_method == "breadth" -%}
+I want you to act as a Prompt Creator.
```
davidberenstein1957, line 2: Feel free to open a PR for this typo in the WizardLM repo

src/distilabel/tasks/_templates/evol-instruct.jinja2
```diff
@@ -0,0 +1,33 @@
...
+Your goal is to draw inspiration from the #Given Prompt# to create a brand new prompt.
+This new prompt should belong to the same domain as the #Given Prompt# but be even more rare.
```
davidberenstein1957, line 4: I would say "but must be even more rare" even more rare. -TYPO

src/distilabel/tasks/text_generation/evol_instruct.py
```diff
@@ -0,0 +1,93 @@
...
+
+@dataclass
+class EvolInstructTask(TextGenerationTask):
+    """A `TextGenerationTask` following the `EvolInstruct` specification for building the prompts.
+
+    Args:
+        system_prompt (str, optional): the system prompt to be used. Not defined for this task.
+
+    Notes:
+        - The paper mentions a parameter *M* for the number of evolutions. This behavior can be achieved dealing
```
davidberenstein1957, line 39: can't we add this as argument to the task itself?

src/distilabel/tasks/text_generation/evol_instruct.py
```diff
@@ -0,0 +1,93 @@
...
+        with the dataset directly, generating copies of the inputs. Take a look at
+        [examples/pipeline-evol-instruct-alpaca.py](https://github.com/argilla-io/distilabel/tree/main/examples/pipeline-evol-instruct-alpaca.py)
+        script for an example.
+        - The original implementation includes an elimination step, which is not implemented currently.
```
davidberenstein1957, line 43: should we create a follow-up issue to support this?

davidberenstein1957: I believe points 2-4 can be implemented already. We might have `validation` functions?

1. The evolved instruction does not provide any information gain compared to the original one.
We use ChatGPT to make this determination, details please refer to Appendix G.
2.The evolved instruction makes it difficult for the LLM to generate a response. We found that
when the generated response contains “sorry” and is relatively short in length (i.e., less than
80 words), it often indicates that the LLM struggles to respond to the evolved instruction.
So we can use this rule to make a judgment.
3. The response generated by the LLM only contains punctuation and stop words.
4. The evolved instruction obviously copies some words from the evolving prompt, such as
“given prompt”, “rewritten prompt”, “#Rewritten Prompt#”, etc

davidberenstein1957: potentially 1 can be implemented in a more cumbersome way.

src/distilabel/tasks/text_generation/evol_instruct.py
davidberenstein1957: Similar to the mention below, I would say the value of this task is hidden in the automated evolution of the dataset.

plaguss: Reviewing the code currently we cannot do what the parameter M implies. For example in the OpenAI implementation, it already has a `num_generations` argument which calls the API that amount of times, but currently, modifying this argument we don't call the task `generate_prompt` again. We could review the implementation, but I wanted to keep it as is for the moment

src/distilabel/tasks/text_generation/evol_instruct.py
plaguss: Talking with @dvsrepo we thought it wouldn't add that much value. The first one currently is not doable, could be nice to review it upon refactoring of the `Pipeline`. For the points 2-4 I can add some of the validations and see how they work

davidberenstein1957: I afree with 1 being too complext but 2-4 seem like very basic checks that don't rely on much additional checks and that would also reduce cost and inference time.

src/distilabel/tasks/text_generation/evol_instruct.py
davidberenstein1957: Yes I see. I'm not too familiar with the API yet but implementing this would be cumbersome.

davidberenstein1957: I think we can emphasize the need for multi-evolutions a bit more clearly in the docs instead @plaguss

plaguss: Rewritten at the docstrings for the task.

src/distilabel/tasks/text_generation/evol_instruct.py
plaguss: Added the implementation for the elimination step

src/distilabel/tasks/text_generation/evol_instruct.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # 2) The evolved instruction makes it difficult for the LLM to generate a response.
+        if "sorry" in output.lower() and len(output.split(" ")) < 80:
+            return
+
+        # 3) The output only contains punctuation and stop words
+        if (
+            len(set(output).intersection(string.punctuation)) / len(set(output))
+        ) == 1.0:
+            logger.info(
```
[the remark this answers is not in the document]
davidberenstein1957: We might be able to use the CURL download approach to get a list of words or we should just skip it.?

src/distilabel/tasks/text_generation/evol_instruct.py
```diff
@@ -0,0 +1,155 @@
...
+            formatted_prompt=self.template.render(**render_kwargs),
+        )
+
+    @property
+    def output_args_names(self) -> List[str]:
+        return ["instruction"]
+
+    def _elimination_evolving(self, output: str) -> Optional[str]:
+        """Performs the elimination step of the Evol-Instruct task, steps 2-4 in the paper:
+
```
davidberenstein1957, line 109: Maybe we should still outline step 1? and add something [NOT IMPLEMETED] 1. blabla bla?

src/distilabel/tasks/text_generation/evol_instruct.py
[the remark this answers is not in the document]
plaguss: I forget about adding the stop words and only checked the punctuation signs. Originally was trying to set a threshold instead of checking for every character being a punctuation sign, but it implies dealing with that threshold... so I let it like that. I can try going with the curl download instead of adding a dependency for this

src/distilabel/tasks/text_generation/evol_instruct.py
```diff
@@ -0,0 +1,155 @@
...
+        2. The evolved instruction makes it difficult for the LLM to generate a response. We found that
+        when the generated response contains “sorry” and is relatively short in length (i.e., less than
+        80 words), it often indicates that the LLM struggles to respond to the evolved instruction.
+        So we can use this rule to make a judgment.
+        3. The response generated by the LLM only contains punctuation and stop words.
```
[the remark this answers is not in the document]
plaguss: Mentioned in other of the comments, I forgot adding those and went just with the punctuation signs

src/distilabel/tasks/text_generation/evol_instruct.py
```diff
@@ -0,0 +1,155 @@
...
+        “given prompt”, “rewritten prompt”, “#Rewritten Prompt#”, etc.
+        """
+        if output == "":
+            return
+
+        # 2) The evolved instruction makes it difficult for the LLM to generate a response.
+        if "sorry" in output.lower() and len(output.split(" ")) < 80:
+            return
+
+        # 3) The output only contains punctuation and stop words
```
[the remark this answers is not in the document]
plaguss: What dependency do you refer to?

src/distilabel/tasks/text_generation/evol_instruct.py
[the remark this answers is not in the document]
plaguss: Maybe I can just add this list of [stopwords](https://github.com/stopwords-iso/stopwords-en/blob/master/stopwords-en.txt) internally and avoid the download, let me know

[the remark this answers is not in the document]
davidberenstein1957: They don't really mention a threshold in the paper right?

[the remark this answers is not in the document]
plaguss: no, wil do just all the punctuation signs instead of the threshold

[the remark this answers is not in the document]
plaguss: Updated

src/distilabel/tasks/text_generation/evol_instruct.py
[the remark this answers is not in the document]
davidberenstein1957: If we already by chance were downloading `nltk` we might have been able to re-use their logic. Sadly, we don't

src/distilabel/tasks/text_generation/evol_instruct.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def parse_output(self, output: str) -> Dict[str, List[str]]:
+        """Parses the output of the model into the desired format, applying the elimination step for bad generations.
+
+        Args:
+            output (str): the output of the model.
+
+        Note:
+            The eliminatin step is applied to the output, but only steps 2-4 in the paper are implemented.
+            Refer to point 3.2, Elimination Evolving section in the original paper for more information on the
```
davidberenstein1957, line 188: 3.2 of what? should we not mention the Arxiv reference? And perhaps also do that in the main docstring?

davidberenstein1957 (approved):

src/distilabel/tasks/text_generation/evol_instruct.py
plaguss: Updated with the explicit reference to the paper

docs/technical-reference/tasks.md
```diff
@@ -46,14 +46,41 @@ For the API reference visit [TextGenerationTask][distilabel.tasks.text_generatio
...
+
+```python
+--8<-- "docs/snippets/technical-reference/tasks/generic_openai_evol_instruct.py"
+```
+
+You can take a look at a [sample dataset](https://huggingface.co/datasets/argilla/distilabel-sample-evol-instruct?row=19) generated using the script the following script: [examples/pipeline-evol-instruct-alpaca.py](../../examples/pipeline-evol-instruct-alpaca.py).
+
+!!! note
+    The original definition of `EvolInstruct` considers an elimination evolving step with different
+    situations to remove the responses considered as failures. Section 3.2, *Elimination Evolving* in [WizardLM paper](https://arxiv.org/abs/2304.12244) shows these steps. We have implemented steps 2-4 as part of this task, but not step one. If the user wants to tackle this, an example can be seen in the following script: [examples/pipeline-openai-wizardl-equal-prompts.py](../../examples/pipeline-evol-instruct-alpaca.py).
```
davidberenstein1957, line 78: the wrong file is referenced here

davidberenstein1957: also mention more explicitly that it shows how to implement evaluation step 1 using distilabel please

PR #249, opened 2024-01-15
docs: Update links to arxiv landing pages rather than PDFs

davanstrien:
Sorry, it's a very pedantic change, but I think it's nicer to point to the Arxiv landing page rather than the PDFs, especially since Arxiv is slowly adding HTML rendering for papers.


davidberenstein1957 (approved):
Hi @davanstrien, I agree. Thanks for the contribution.

Issue #246, opened 2024-01-15
[FEATURE] align argument usage across different llm providers


davidberenstein1957:
**Is your feature request related to a problem? Please describe.**
Across different LLM providers we seem to be using different ways of configuration.
- We swap the order of arguments.: sometimes `model` is the 1st argument and other times `task`, additionally generation arguments are ordered differently
- We have different naming: sometimes we use `model` sometimes `vllm`, sometimes `openai_api_key` sometimes `api_key`

**Describe the solution you'd like**
I would like to align the naming to `model` and `api_key`. Additionally, I would like to see `client` and `api_key` as last arguments because they are non-default and only apply to certain LLM integrations

**Describe alternatives you've considered**
Using the mixed config we have now.

**Additional context**
N.A.

alvarobartt:
Hi @davidberenstein1957 the main reason behind this is that `task` is always needed and mandatory while e.g. `model` is not mandatory in all the cases as we already provide a default one. Additionally, w.r.t. the rest of the args within the `LLM` those are specific to each LLM engine in most of the scenarios, so would be hard to align, and even in some cases, nonsense. Also providing `generation_kwargs` as a mapping hides visibility from those LLM-specific args, we had that conversation initially when developing it, but at the end we decided to go with `kwargs` even those were differing from one LLM engine to another.

davidberenstein1957:
@alvarobartt I agree. For me it was not clear this was the reasoning behind it. Then we should also change the communication here: https://distilabel.argilla.io/latest/technical-reference/llms/#llm and the various `__rich_repr__`-methods. Besides that, the ordering issue w.r.t. variable introduction persists IMO.

PR #250, opened 2024-01-15
feat: add support `ollama` api

davidberenstein1957:
In this PR I've included basic support for [ollama](https://ollama.ai/)

Some design choices:

- [x] use builtin `urllib` library. we might replace this with `requests` to also allow for passing a `client`
- [x] retry 5 times with n+1 backoff in case of 500 (server errors) as fixed config for some robustness but to avoid spending a lot of time on this
- [x] default to `http://localhost:11434` but overwrite with `OLLAMA_HOST` env var as in Ollama docs
- [x] [generation kwargs](https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values) don't align with other standards (maybe we can forcefully set them to deployment config when running `ollama run model`?)

Things to note:
- we use the `api/chat` endpoint
- we can't do `n_generations` in one request so we make multiple


src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,157 @@
...
+
+        try:
+            # Send the request
+            with request.urlopen(req) as response:
+                # Check if the request was successful (status code 200)
+                if response.getcode() == 200:
+                    # Parse and return the response JSON
+                    return json.loads(response.read().decode("utf-8"))
+                elif response.getcode() >= 500 and n < 5:
+                    # If the request failed, try again with backoff
```
plaguss, line 102: We already have backoff implemented with [tenacity](https://github.com/jd/tenacity), you can take a look at the inference endpoints from HF [here](https://github.com/argilla-io/distilabel/blob/main/src/distilabel/llm/huggingface/inference_endpoints.py#L193-L202) for reference.

plaguss (commented):
Nice addition!! For when you tackle the docs, it would be nice to include an example of how to setup ollama so it is as contained as possible

davidberenstein1957:
Difficulty w.r.t. `__init__` args was first defined here https://github.com/argilla-io/distilabel/issues/246.

plaguss (commented):
Nice! Could you add some tests like the ones [here](https://github.com/argilla-io/distilabel/tree/main/tests/llm)? After that we should have a new cool integration 😃

PR #260, opened 2024-01-16
feat: Fixing progress bar and ETA

### LINKED ISSUE #258: [FEATURE] Further changes on ETA for progress bars
There are still some additional changes to be done to the ETA for progress bars:

- [ ] Check OpenAI's future
- [ ] has_labeller and has_generator are not working

ignacioct:
Closes #258

### DIFF
--- a/src/distilabel/progress_bar.py
+++ b/src/distilabel/progress_bar.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from functools import partial, wraps
+from functools import wraps
 from typing import Any, Callable, Tuple, TypeVar, Union
 
 from rich.progress import (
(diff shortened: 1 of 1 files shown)


davidberenstein1957:
Thanks @ignacioct, it looks much cleaner.

ignacioct:
> Thanks @ignacioct, it looks much cleaner.

@davidberenstein1957 Yes, was looking for a way of not initializing anything unnecessary. I'm still doing some tests, there are cases in which it does not show any ETA, just elapsed times at the end, and I'm unsure if it's my doing or the OpenAI's issue.

davidberenstein1957:
I think it will compute the ETA based on leveraging out returned values and times in between so maybe your runs are too short to do a proper computation?

ignacioct:
> @ignacioct I think it will compute the ETA based on leveraging out returned values and times in between so maybe your runs are too short to do a proper computation?

Could you share the script you used to test it? Thanks!

davidberenstein1957:
I used the pipeline-llamacpp-and-openai.py and just replaced both LLMs with OpenAI ones :)

PR #262, opened 2024-01-17
feat: 230 feature add sentence transformers support for the to argilla method

davidberenstein1957:
I added `sentence-transformers` support to the `to_argilla` methods.

Things to note:
- I added an `add_vectors_to_argilla_dataset` to the `Task` base class
- I added a `vector_strategy` argument to the `to_argilla` methods. This can either be a bool or a `SentenceTransformersExtractor` and default to True. If True, it uses the defaults. if False, it does nothing, if SentenceTransformersExtractor it will use a custom-initialized extractor.
- I added `sentence-transformers` to the `argilla` extras to avoid having more splits in this and to align with using this as default.
- I bumped the argilla version to `1.22.0` because this is required for the `SentenceTransformersExtractor`

WIP: 
- docs and examples


src/distilabel/tasks/base.py
```diff
@@ -119,6 +128,44 @@ def to_argilla_record(
...
+
+        elif not _ARGILLA_AVAILABLE and vector_strategy:
+            warnings.warn(
+                "An error occurred while adding vectors to the dataset: "
+                "The `argilla`/`sentence-transformers` packages are not installed or the installed version is not compatible with the"
+                " required version. If you want to add vectors to your dataset, please run `pip install 'distilabel[vectors]'`.",
+                stacklevel=2,
+            )
+        else:
+            pass
```
plaguss, line 166:
```suggestion

```

plaguss: Maybe we can remove this else: pass if we have to take into account argilla for this

src/distilabel/tasks/text_generation/self_instruct.py
```diff
@@ -149,11 +156,17 @@ def to_argilla_dataset(self, dataset_row: Dict[str, Any]) -> "FeedbackDataset":
...
         # Then we just return the `FeedbackDataset` with the fields, questions, and metadata properties
         # defined above.
-        return rg.FeedbackDataset(
+        dataset = rg.FeedbackDataset(
             fields=fields,
             questions=questions,  # type: ignore
             metadata_properties=metadata_properties,  # Note that these are always optional
         )
+        dataset: (
+            FeedbackRecord | List[FeedbackRecord] | FeedbackDataset
```
plaguss, line 165: Does this `Union` format works for python 3.8? I'm not sure, but if that's the case would love to update it all around the codebase

plaguss (commented):
Some small comments, this will be so useful! Some pictures of the UI would be nice once we add the docs, just to keep it in mind

src/distilabel/tasks/text_generation/self_instruct.py
davidberenstein1957: good catch

src/distilabel/tasks/base.py
davidberenstein1957: I agree

src/distilabel/dataset.py
```diff
@@ -93,8 +113,64 @@ def to_argilla(self) -> "FeedbackDataset":
...
                     stacklevel=2,
                 )
+
+        # set columns to all input and output columns for the task
+        if dataset_columns is None:
+            dataset_columns = getattr(self.task, "input_args_names", []) + getattr(
+                self.task, "output_args_names", []
+            )
+        # get the first 5 that align with column selection + f"{column_name}_idx"
+        selected_fields = []
```
plaguss, line 123:
```suggestion
```

src/distilabel/tasks/base.py
```diff
@@ -18,13 +18,16 @@
...
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, Dict, Generator, List, Literal, Union
 
 from jinja2 import Template
 
 from distilabel.tasks.prompt import Prompt
+from distilabel.utils.imports import _ARGILLA_AVAILABLE
+
+if _ARGILLA_AVAILABLE:
+    pass
```
plaguss, line 30:
```suggestion
```

plaguss: has no effect currently

plaguss (commented):
Some small notes, looks nice, will be really helpful 😄

tests/test_dataset.py
```diff
@@ -129,6 +138,7 @@ def test_dataset_save_to_disk(custom_dataset):
         assert (ds_name / "task.pkl").is_file()
```
plaguss, line 138: Could you update the "task.pkl" to the proper variable name? That's my bad... I'm updating this in a different PR, and the change will ensure this test still works as expected: 
```python
from distilabel.utils.dataset import TASK_FILE_NAME
```

davidberenstein1957:
@plaguss I now added a `dataset_columns` option, which allows either `datasets` columns or `FeedbackDataset` fields to select. What do you think of the naming?

plaguss (approved):
The naming looks fine to me 😄

Issue #247, opened 2024-01-15
[DOCS] add a tutorial on "clean a DPO/preference dataset with distilabel"


davidberenstein1957:
## Which page or section is this issue related to?
tutorials

## What are you documenting, or what change are you making in the documentation?
@dvsrepo made a wonderful effort curating the [intel-orca pairs](https://huggingface.co/datasets/argilla/distilabel-intel-orca-dpo-pairs) using distilabel but there is currently no tutorial for it. 

@ignacioct you should be able to copy-paste most steps and the outline from [the HF model card](https://huggingface.co/datasets/argilla/distilabel-intel-orca-dpo-pairs)

PR #264, opened 2024-01-17
Add new functionality to binarize preference datasets directly from distilabel

### LINKED ISSUE #263: [FEATURE] Add functionality to binarize datasets for fine-tuning directly from distilabel
We need helper functions to transform our preference datasets to the format expected for trainers like [DPOTrainer](https://huggingface.co/docs/trl/main/en/dpo_trainer).

Functions to transform the dataset from a `Pipeline` in a format suitable for fine-tuning alignment.

Do it by hand.

**Additional context**

plaguss:
## Description

This PR adds a function to binarize a `CustomDataset` with a `PreferenceTask`:

```python
from distilabel.utils import prepare_dataset
from datasets import load_dataset
from distilabel.tasks import JudgeLMTask
from distilabel.dataset import prepare_dataset

dataset = load_dataset("argilla/distilabel-intel-orca-dpo-pairs", split="train")
# Add the task that originally was used for the labelling
dataset.task = JudgeLMTask()
dataset_binarized_random = prepare_dataset(dataset, strategy="random", keep_ties=True)
```

Closes #263


src/distilabel/utils/dataset.py
```diff
@@ -51,3 +60,215 @@ def load_task_from_disk(path: Path) -> "Task":
     with open(task_path, "rb") as f:
         task = pickle.loads(f.read())
     return task
+
+
+def _binarize_dataset(
+    dataset: "CustomDataset",
+    seed: int = None,
+    strategy: BinarizationStrategies = "random",
+    keep_ties: bool = True,
```
davidberenstein1957, line 69: why did we set `keep_ties` to `True` as default? Intuitively I would say removing them will lead to higher quality data.

src/distilabel/utils/dataset.py
```diff
@@ -51,3 +60,215 @@ def load_task_from_disk(path: Path) -> "Task":
...
+    **kwargs: Any,
+) -> "CustomDataset":
+    """Binarizes a distilabel dataset.
```
davidberenstein1957, line 72: again, some people might be like: "what is binarizing?"

src/distilabel/utils/dataset.py
```diff
@@ -51,3 +60,215 @@ def load_task_from_disk(path: Path) -> "Task":
...
+    Raises:
+        ValueError: If the strategy is not implemented.
+
+    Returns:
+        CustomDataset: Dataset binarized.
+    """
+    rating_column = "rating"
+    responses_column = "generations"
+
+    def binarize_random(example):
```
davidberenstein1957, line 92: there seems to be some code duplication in binarize_random and binarize_worst, perhaps we can combine these a bit more?

davidberenstein1957, line 92: type hints

src/distilabel/utils/dataset.py
```diff
@@ -51,3 +60,215 @@ def load_task_from_disk(path: Path) -> "Task":
...
+            "chosen": chosen_response,
+            "rejected": random_response,
+            "rating_chosen": int(best_rating),
+            "rating_rejected": int(random_rating),
+            "chosen_model": chosen_model,
+            "rejected_model": random_model,
+        }
+        return binarized
+
+    def binarize_worst(example):
```
davidberenstein1957, line 137: time hints?

src/distilabel/utils/dataset.py
```diff
@@ -51,3 +60,215 @@ def load_task_from_disk(path: Path) -> "Task":
...
+            lambda example: example["rating_chosen"] != example["rating_rejected"]
+        )
+    return dataset
+
+
+def prepare_dataset(
+    dataset: "CustomDataset",
+    strategy: BinarizationStrategies = "random",
+    seed: Optional[int] = None,
+    keep_ties: bool = True,
```
davidberenstein1957, line 193: same here w.r.t. keep_ties=True

src/distilabel/utils/dataset.py
plaguss: Wasn't sure what value to set as default, and thought that allowed to keep all the data initially, but yes, I think it makes more sense to remove those by default

src/distilabel/utils/dataset.py
plaguss: sure will take a look

plaguss: they are not that usual in internal functions, and that's the standard example from [HuggingFace's datasets](https://huggingface.co/docs/datasets/process#map).

src/distilabel/utils/dataset.py
plaguss: same as `binarize_random`

src/distilabel/utils/dataset.py
plaguss: Could you both @davidberenstein1957 and @dvsrepo take a look at the new section of the docs for the dataset preparation/binarization? This commit https://github.com/argilla-io/distilabel/pull/264/commits/e4602d4e2c0a68737915cfe6a743ead343091d4e contains the updates.

dvsrepo:
@plaguss it looks good, I don't have cycles to do an in-depth review.

One suggestion: would it be possible to return the chosen and rejected in the OpenAI format, like this?

https://huggingface.co/datasets/HuggingFaceH4/ultrafeedback_binarized?row=0

I think this would make it more interoperable

plaguss:
> @plaguss it looks good, I don't have cycles to do an in-depth review.
> 
> One suggestion: would it be possible to return the chosen and rejected in the OpenAI format, like this?
> 
> https://huggingface.co/datasets/HuggingFaceH4/ultrafeedback_binarized?row=0
> 
> I think this would make it more interoperable

It should be easy, will take a look and update the code

davidberenstein1957 (approved):

davidberenstein1957:
@plaguss can you resolve the merge conflicts. After that you can merge it. @sdiazlor , this might be interesting to include or mention in your tutorial too. https://github.com/argilla-io/distilabel/issues/247

PR #269, opened 2024-01-18
feat: add customizable criteria for query generation in SelfInstructTask

### LINKED ISSUE #255: [FEATURE] Add more flexibilty in construction of the SelfInstructTask template
When using Distilabel recently, I wanted to modify the ` # Criteria for Queries` part of the task template. Since this isn't currently exposed via the `SelfInstructTask` constructor I ended up patching  [self-instruct.jinja2](https://github.com/argilla-io/distilabel/blob/main/src/distilabel/tasks/_templates/self-instruct.jinja2) directly. 

It might be nice to also expose this part of the template for the user to modify? 

At some point, the tasks might be modified so that they don't have much similarity to the original paper. I don't think this is such a problem but another possible approach would be to to add more functionality or documentation for constructing custom tasks.

ignacioct:
Closes #255 suggested by @davanstrien 

I've tried with https://github.com/davanstrien/haiku-dpo/blob/main/01_generate_haiku_prompts.ipynb and works fine. I've updated the documentation with a generic example instead of a Haiku example, to be more consistent with the rest of the snippets, which are very generic. We could do a tutorial with some use cases, like the Haiku, to show off this new feature.

Thanks, @dvsrepo and @plaguss for showing me the way 💡

### DIFF
--- a/src/distilabel/tasks/text_generation/self_instruct.py
+++ b/src/distilabel/tasks/text_generation/self_instruct.py
@@ -67,9 +67,9 @@ class SelfInstructTask(TextGenerationTask):
 
     criteria_for_query_generation: str = (
         "Incorporate a diverse range of verbs, avoiding repetition. "
-        "Ensure queries are compatible with AI model's text generation functions and are limited to 1-2 sentences."
-        "Design queries to be self-contained and standalone."
-        """Blend interrogative (e.g., "What is the significance of x?") and imperative (e.g., "Detail the process of x.") styles."""
+        "Ensure queries are compatible with AI model's text generation functions and are limited to 1-2 sentences. "
+        "Design queries to be self-contained and standalone. "
+        """Blend interrogative (e.g., "What is the significance of x?") and imperative (e.g., "Detail the process of x.") styles. """
     )
 
     __jinja2_template__: str = _SELF_INSTRUCT_TEMPLATE
--- a/src/distilabel/tasks/text_generation/self_instruct.py
+++ b/src/distilabel/tasks/text_generation/self_instruct.py
@@ -61,8 +61,14 @@ class SelfInstructTask(TextGenerationTask):
         " You are given a task description and a set of instructions for how to write the prompts for an"
         " specific AI application."
     )
+
     application_description: str = "AI assistant"
     num_instructions: int = 5
+
+    criteria_for_query_generation = """Incorporate a diverse range of verbs, avoiding repetition.
+    Ensure queries are compatible with AI model's text generation functions and are limited to 1-2 sentences.
+    Design queries to be self-contained and standalone.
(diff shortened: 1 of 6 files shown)


src/distilabel/tasks/text_generation/self_instruct.py
```diff
@@ -61,9 +61,17 @@ class SelfInstructTask(TextGenerationTask):
...
     application_description: str = "AI assistant"
     num_instructions: int = 5
 
+    criteria_for_query_generation: str = (
+        "Incorporate a diverse range of verbs, avoiding repetition. "
+        "Ensure queries are compatible with AI model's text generation functions and are limited to 1-2 sentences. "
+        "Design queries to be self-contained and standalone. "
+        """Blend interrogative (e.g., "What is the significance of x?") and imperative (e.g., "Detail the process of x.") styles. """
+    )
+
```
plaguss, line 74:
```suggestion
        "Incorporate a diverse range of verbs, avoiding repetition.\n"
        "Ensure queries are compatible with AI model's text generation functions and are limited to 1-2 sentences.\n"
        "Design queries to be self-contained and standalone.\n"
        """Blend interrogative (e.g., "What is the significance of x?") and imperative (e.g., "Detail the process of x.") styles.\n"""
    )

```

plaguss: Not really sure of how much it could affect the "\n", but just to keep it it was originally

plaguss (commented):
nice! just the small comment to keep the original template intact by default

davidberenstein1957 (commented):
Looks good.

@ignacioct @plaguss, for me now it is only implicitly possible to change these values but we don't provide any clear insights in the docs or pointers as to how to do that. Also, we need to make a reference to it within the docs.

davidberenstein1957:
@ignacioct maybe you can also add references or mention the jinaj2 templates when you tackle the customising task templates issue w.r.t. the documentation.

ignacioct:
> Looks good.
> 
> @ignacioct @plaguss, for me now it is only implicitly possible to change these values but we don't provide any clear insights in the docs or pointers as to how to do that. Also, we need to make a reference to it within the docs.

@davidberenstein1957 I've added the docstring and docs changes, please tell me if we need further documentation

plaguss:
Just some extra addition to the docs and done

ignacioct:
Ready for final review, and we can push @plaguss @davidberenstein1957

davidberenstein1957:
@ignacioct , can you add the custom task snippet(s) in another PR?

plaguss (approved):

plaguss:
> @ignacioct , can you add the custom task snippet(s) in another PR?

Do you mean to remove the case of the `SelfInstructTask`? I think it makes sense here as a case of customising that one specifically

davidberenstein1957 (approved):

PR #271, opened 2024-01-18
feat: 231 feature add text descriptives support to the to argilla methods

davidberenstein1957:
should be merged after https://github.com/argilla-io/distilabel/pull/262 introducing `sentence-transfomers`


pyproject.toml
plaguss, line 1: Maybe this version should be the same in the `imports.py` file?

plaguss (commented):
Other than the versions looks perfect!

PR #273, opened 2024-01-19
Update README.md - load_from_disk

7flash:
I think it's important to know how to load dataset from local file after it's been generated, particularly in case if remote uploading failed, user wants to avoid regenerating dataset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,15 @@ pipeline = pipeline("preference", "instruction-following", generator=generator)
 dataset = pipeline.generate(dataset)
 ```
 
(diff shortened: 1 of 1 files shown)


plaguss:
Hi @7flash! This is in the docs [here](https://distilabel.argilla.io/latest/technical-reference/pipeline/#dataset-checkpoints), do you think is it clear? Maybe it should be more accessible

davidberenstein1957:
thank you for the PR @7flash , any other help is appreciated and we will discuss some things during our meeting next week.

PR #275, opened 2024-01-19
Custom Task Documentation

### LINKED ISSUE #256: [DOCS] add example on defining a custom `Task`

https://distilabel.argilla.io/latest/technical-reference/tasks/ does not contain a proper example.

## What are you documenting, or what change are you making in the documentation?

The tutorial @ignacioct has been working on might be a nice base to do so.

ignacioct:
Closes #256


ignacioct:
@davidberenstein1957 I've tested the tab structure in the TextGenerationTask, but I've encountered two little problems that make me a little sceptical:

- We loose the quick access in the tree of content, so in theory users can miss this content if they just look at that
- UltraFeedback already has a tab structure, so I cannot nest two tab structures

Screenshot 2024-01-19 at 12 37 19

plaguss:
> @davidberenstein1957 I've tested the tab structure in the TextGenerationTask, but I've encountered two little problems that make me a little sceptical:
> 
> * We loose the quick access in the tree of content, so in theory users can miss this content if they just look at that
> * UltraFeedback already has a tab structure, so I cannot nest two tab structures
> 

Maybe it's better to add an extra section under `TextClassificationTask` as if it was a different case (SelfInstructTask, EvolInstructTask, TextGenerationTask, CustomTextGenerationTask...), and add there the steps along with some example. I think that should be enough

ignacioct:
Looking for feedback on this one @davidberenstein1957 @plaguss, specially on the custom critique and custom preference, cause I never used those tasks and I'm not very familiar with them :)

docs/technical-reference/tasks.md
```diff
@@ -66,9 +64,9 @@ You can personalize the way in which your SelfInstructTask behaves by changing t
 
 Let's see an example of how to customise a SelfInstructTask to create Haikus in the snippet below. You can take a look at this dataset as an example of a [Haiku DPO dataset](https://huggingface.co/datasets/davanstrien/haiku_dpo).
 
-```python
+````python
 --8<-- "docs/snippets/technical-reference/tasks/custom_task_selfinstruct_haikus.py"
-```
+    ```
```
plaguss, line 69: Does this render properly?

docs/technical-reference/tasks.md
```diff
@@ -83,7 +81,7 @@ Use this `Task` to build more complete and complex datasets starting from simple
 
 ```python
 --8<-- "docs/snippets/technical-reference/tasks/generic_openai_evol_instruct.py"
-```
+````
```
plaguss, line 84: Same here, it seems there is one extra "`" character

plaguss (changes requested):
Good job @ignacioct! I think we need to review some of the cases though.

plaguss:
For the `TextGenerationTasks` I think it's worth keeping both the OSS-Instruct and the WizardLM, but the `Custom...` doesn't add too much. Similar with the `Critique` and `Preference`, I think those might confuse a bit.
So maybe we can modify the approach. For the `TextGenerationTasks`, using those ones, tell how we can customise the tasks, for example in these cases we are implementing ... (add the paper references and highlight the methods that need to be implemented and what they do). After that let's review the customisation of the `Critique` and `Preference` tasks 😄

davidberenstein1957:
@ignacioct this note is ill-rendered can you tackle that in this PR too? 

https://distilabel.argilla.io/latest/technical-reference/tasks/#evolinstructtask

docs/snippets/technical-reference/tasks/custom_task_selfinstruct_haikus_docs.py
```diff
@@ -0,0 +1,33 @@
+from dataclasses import dataclass
```
davidberenstein1957, line 1: why the new file and not the old one?

docs/technical-reference/tasks.md
```diff
@@ -67,7 +65,7 @@ You can personalize the way in which your SelfInstructTask behaves by changing t
 Let's see an example of how to customise a SelfInstructTask to create Haikus in the snippet below. You can take a look at this dataset as an example of a [Haiku DPO dataset](https://huggingface.co/datasets/davanstrien/haiku_dpo).
 
 ```python
---8<-- "docs/snippets/technical-reference/tasks/custom_task_selfinstruct_haikus.py"
+--8<-- "docs/snippets/technical-reference/tasks/custom_task_selfinstruct_haikus_docs.py"
```
davidberenstein1957, line 68: @I think it might be cleaner to add this one to the rest of the examples?

davidberenstein1957: with this, I mean the haiku example

ignacioct: It was just a name change, I can revert it, but thought it would be useful to state that this snippet is for the docs

docs/snippets/technical-reference/tasks/custom_task_selfinstruct_haikus_docs.py
ignacioct: It was just a name change, I can revert it, but thought it would be useful to state that this snippet is for the docs

docs/technical-reference/tasks.md
ignacioct: so, we leave the SelfInstructTask with no example and we add the Haiku example to all the example of CustomTasks?

ignacioct:
> @ignacioct this note is ill-rendered can you tackle that in this PR too?
> 
> https://distilabel.argilla.io/latest/technical-reference/tasks/#evolinstructtask

Done :)

PR #288, opened 2024-01-22
Add new serialization strategy

### LINKED ISSUE #261: [FEATURE] Improve serialization strategy for tasks
As discussed internally we have to improve the serialization/deserialization of the components in to prepare for further enhancements, starting by the `Task` classes.

A serialization/deserialization definition for the tasks, safe and preferably text based

Continue with the dill based definition.

**Additional context**
Once this is defined we should update the push_to_hub and load_dataset to deal with the serialized task out of the box.

plaguss:
## Description

This PR adds a new module `distilabel.utils.serialization` to allow custom serialization of the tasks (to be extended to the different distilabel components in the future). When we save/load from disk, we will have a `task-distilabel.json` instead of `task.pkl`.

- This definition will replace the current one based in `pickle`, making it safer to move the files around.
- Currently the definition is completely custom, but we could use the pydantic one if we see fit.

Sample task from `UltraFeedbackTask`:

```json
{
  "system_prompt": "You are an expert prompt writer, writing the best and most diverse prompts for a variety of tasks. You are given a task description and a set of instructions for how to write the prompts for an specific AI application.",
  "principles": {
    "harmlessness": [
      "The assistant should not provide answers that could cause physical, psychological, or social harm to any group of people; damage to or loss of equipment or property; damage to the environment; or harm to institutions or resources necessary to human wellbeing. Feel free to refuse to answer questions that could cause the above harm.",
     "..."
    ],
    "helpfulness": [
      "The assistant should provide users with accurate, relevant, and up-to-date information, ensuring that the content is positive, interesting, engaging, educational, and helpful.",
      "As an AI assistan..."
  "principles_distribution": null,
  "application_description": "AI assistant",
  "num_instructions": 5,
  "criteria_for_query_generation": "Incorporate a diverse range of verbs, avoiding repetition.\nEnsure queries are compatible with AI model's text generation functions and are limited to 1-2 sentences.\nDesign queries to be self-contained and standalone.\nBlend interrogative (e.g., \"What is the significance of x?\") and imperative (e.g., \"Detail the process of x.\") styles.",
  "__jinja2_template__": "...",
  "__type_info__": {
    "module": "distilabel.tasks.text_generation.self_instruct",
    "name": "SelfInstructTask"
}
```

Closes #261.


src/distilabel/dataset.py
```diff
@@ -21,8 +21,8 @@
 from datasets import Dataset
 
 from distilabel.utils.argilla import infer_field_from_dataset_columns
-from distilabel.utils.dataset import load_task_from_disk, save_task_to_disk
 from distilabel.utils.imports import _ARGILLA_AVAILABLE
+from distilabel.utils.serialization import load_task_from_disk
```
davidberenstein1957, line 25: shouldn't we be able to replace `load_task_from_disk` too?

src/distilabel/utils/serialization.py
```diff
@@ -0,0 +1,154 @@
...
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Dict, Generic, Optional, TypeVar
+
+if TYPE_CHECKING:
+    from distilabel.tasks.base import Task
+
+
+T = TypeVar("T")
+
+TASK_FILE_NAME = "task-distilabel.json"
```
davidberenstein1957, line 28: "distilabel-task.json" might be nicer?

src/distilabel/utils/serialization.py
```diff
@@ -0,0 +1,154 @@
...
+            Dict[str, Any]: Serializable content of the class.
+        """
+        _dict = asdict(self)
+        _dict["__type_info__"] = {
+            "module": type(self).__module__,
+            "name": type(self).__name__,
+        }
+        return _dict
+
+    def save(self, path: Optional[os.PathLike] = None) -> None:
```
davidberenstein1957, line 118: should we add some checks to llow creating the path if not present?

src/distilabel/utils/serialization.py
```diff
@@ -0,0 +1,154 @@
...
+        """Writes to a file the content.
+
+        Args:
+            path (Optional[os.PathLike], optional):
+                Filename of the task. If a folder is given, will create the task
+                inside. If None is given, the file will be created at the current
+                working directory. Defaults to None.
+        """
+        if path is None:
+            path = Path.cwd() / TASK_FILE_NAME
```
davidberenstein1957, line 128: in what scenario would having the path be optional be relevant?

src/distilabel/utils/serialization.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            path = Path(path)
+            if path.suffix == "":
+                # If the path has no suffix, assume the user just wants a folder to write the task
+                path = path / TASK_FILE_NAME
+        write_json(path, self.dump())
+
+    @classmethod
+    def from_json(cls, template_path: os.PathLike) -> Generic[T]:
```
davidberenstein1957, line 137: should we just keep the arg name as `path`? to keep it as uniform as possible_

tests/tasks/test_serialization.py
```diff
@@ -0,0 +1,351 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tempfile
+from pathlib import Path
+
+import pytest
```
davidberenstein1957, line 18: maybe we should dd tests for the base task class too in case people write custom ones?

src/distilabel/utils/serialization.py
plaguss: there are checks in place, isn't it?
```python
    def save(self, path: Optional[os.PathLike] = None) -> None:
        ...
        if path is None:
            path = Path.cwd() / TASK_FILE_NAME
        else:
            path = Path(path)
            if path.suffix == "":
                # If the path has no suffix, assume the user just wants a folder to write the task
                path = path / TASK_FILE_NAME
```

src/distilabel/utils/serialization.py
plaguss: you're right, will update:
```python
    def save(self, path: Optional[os.PathLike] = Path.cwd() / TASK_FILE_NAME) -> None:
        ...
        path = Path(path)
        if path.suffix == "":
            # If the path has no suffix, assume the user just wants a folder to write the task
            path = path / TASK_FILE_NAME
```

src/distilabel/utils/serialization.py
davidberenstein1957: I mean if the path does not exists but probably we'll also see that during the writing

tests/tasks/test_serialization.py
plaguss: Do you mean the `Task`? that shouldn't be instantiated directly, if someone does that I guess it's at their own risk

src/distilabel/dataset.py
plaguss: what do you mean exactly?

davidberenstein1957: We call `self.task.save` but though we might also add `load_task_from_disk` in that `load_from_disk`functon but now see it is inheriting from `Dataset` so that does not make sense

plaguss: will review this

plaguss: Yes those two are different, one for the dataset which we don't really deal with it ourselves, and the task which is saved as a json file.

plaguss:
This still need a change to be ready

PR #293, opened 2024-01-24
Add `MistralAILLM`

### LINKED ISSUE #153: [FEATURE] add Mistral LLM (using their API)
Mistral API models can be very useful for both generation and labelling tasks

plaguss:
## Description

This PR integrates [Mistral AI API](https://docs.mistral.ai/) as a new LLM.

The new `MistralAILLM` is a wrapper over their [python client](https://docs.mistral.ai/platform/client/).

Closes #153


pyproject.toml
```diff
@@ -42,6 +42,7 @@ vertexai = ["google-cloud-aiplatform >= 1.38.0"]
 together = ["together"]
 argilla = ["argilla >= 1.18.0"]
 tests = ["pytest >= 7.4.0"]
+mistralai = ["mistralai"]
```
davidberenstein1957, line 45: should we add a more specific version?

src/distilabel/llm/mistralai.py
```diff
@@ -0,0 +1,157 @@
...
+    from distilabel.tasks.prompt import SupportedFormats
+
+logger = get_logger()
+
+
+class MistralAILLM(LLM):
+    def __init__(
+        self,
+        task: "Task",
+        model: str = "mistral-tiny",
```
davidberenstein1957, line 40: what models did we choose for for example OpenAI? maybe we need to force them to choose a model more specificically

src/distilabel/llm/mistralai.py
```diff
@@ -0,0 +1,157 @@
...
+            )
+
+        self.max_tokens = max_tokens
+        self.temperature = temperature
+        self.top_p = top_p
+        self.seed = seed
+        self.safe_prompt = safe_prompt
+
+        # If the api key is not given, try to get it from the environment
+        if not api_key:
```
davidberenstein1957, line 71: can`t we write this easier?

src/distilabel/llm/mistralai.py
```diff
@@ -0,0 +1,157 @@
...
+        # Explicitly write the default parameters of the model
+        self.client = client or MistralClient(
+            api_key=api_key, max_retries=5, timeout=120
+        )
+        assert (
+            model in self.available_models
+        ), f"Provided `model` is not available in MistralAI, available models are {self.available_models}"
+        self.model = model
+
+    def __rich_repr__(self) -> Generator[Any, None, None]:
```
davidberenstein1957, line 82: we have a lot more params defined above

src/distilabel/llm/mistralai.py
```diff
@@ -0,0 +1,157 @@
...
+                "top_p": self.top_p,
+                "seed": self.seed,
+                "safe_prompt": self.safe_prompt,
+            },
+        )
+
+    @cached_property
+    def available_models(self) -> List[str]:
+        """Returns the list of available models in MistralAI."""
+        return [model.id for model in self.client.list_models().data]
```
davidberenstein1957, line 98: don't models also have a name/version etc like OpenAI?

src/distilabel/llm/mistralai.py
```diff
@@ -0,0 +1,157 @@
...
+        Returns:
+            List[List[LLMOutput]]: the generated outputs.
+        """
+        prompts = self._generate_prompts(inputs, default_format="openai")
+        # The mistralai format is the same as openai, but needs to be converted to mistralai's ChatMessage (pydantic model)
+        prompts = [[ChatMessage(**p) for p in prompt] for prompt in prompts]
+        outputs = []
+        for prompt in prompts:
+            responses = []
+            for _ in range(num_generations):
```
davidberenstein1957, line 126: don't we want to wrap this in retries etc or is this done in the client init?

src/distilabel/llm/mistralai.py
```diff
@@ -0,0 +1,157 @@
...
+                    top_p=self.top_p,
+                    seed=self.seed,
+                    safe_prompt=self.safe_prompt,
+                )
+                responses.append(chat_completion_response)
+
+            output = []
+            for response in responses:
+                chat_completion = response.choices[0]
+                try:
```
davidberenstein1957, line 141: I believe this might be missing in some other implementations?

src/distilabel/utils/imports.py
```diff
@@ -115,3 +115,4 @@ def _check_package_is_available(
     "google-generativeai", min_version="0.3.2", greater_or_equal=True
 )
 _TOGETHER_AVAILABLE = _check_package_is_available("together")
+_MISTRALAI_AVAILABLE = _check_package_is_available("mistralai")
```
davidberenstein1957, line 118: versioning? Perhaps it is better to not do it but just to check.

tests/llm/test_mistralai.py
```diff
@@ -0,0 +1,136 @@
...
+system_prompt = ""
+
+
+@mock.patch("distilabel.llm.mistralai.MistralClient")
+@mock.patch(
+    "distilabel.llm.mistralai.MistralAILLM.available_models",
+    [MODEL_NAME],
+)
+class TestAnyscaleLLM:
+    def test_llm_anyscale(self, mock_mistralai):
```
davidberenstein1957, line 40: naming

src/distilabel/llm/mistralai.py
plaguss: Those are the common ones we've defined in other LLMs, no problem redefining what should we put there to make it common across the LLMs, but maybe it's better to tackle that once we refactor the LLM parameters

src/distilabel/llm/mistralai.py
plaguss: Doesn't seem so: https://docs.mistral.ai/platform/endpoints/#generative-endpoints

src/distilabel/llm/mistralai.py
plaguss: will take a look, but maybe this is a special behaviour of this API, it doesn't deal with `num_generations` in the same way as OpenAI does, so it's easier to grab that choice

src/distilabel/llm/mistralai.py
davidberenstein1957: perhaps the `mistral-small` is more aligned with `gpt-3.5-turbo` performance? But I'll let you choose.

src/distilabel/utils/imports.py
plaguss: Will make it check the same version from pyproject.toml

plaguss:
There's a version conflict with httpx between `argilla` and `mistralai` but works fine when testing that locally. Not sure of the best way to solve this

src/distilabel/llm/mistralai.py
plaguss: yeah that makes sense

src/distilabel/llm/mistralai.py
plaguss: They deal with this themselves, the internal [request mechanism](https://github.com/mistralai/client-python/blob/main/src/mistralai/client.py#L48)

plaguss:
Related to https://github.com/argilla-io/argilla/pull/4531

pyproject.toml
```diff
@@ -42,6 +42,7 @@ vertexai = ["google-cloud-aiplatform >= 1.38.0"]
 together = ["together"]
 argilla = ["argilla >= 1.22.0", "sentence-transformers >= 2.0.0", "textdescriptives >= 2.0.0"]
```
davidberenstein1957, line 43: Shouldn't we pin Argilla to a higher version here?

plaguss: Is there a higher version than `argilla>=1.22` right now?

davidberenstein1957: No, I mean when we merge the `httpx` bumped version

plaguss:
We should be ready once we have https://github.com/argilla-io/argilla/pull/4531 merged

PR #297, opened 2024-01-24
Add functionality to push tasks to the HuggingFace hub and download them automatically.

### LINKED ISSUE #290: [FEATURE] Include the tasks when pushing the `CustomDataset` to HuggingFace hub
With the new serialization strategy it's easier to upload/download the tasks on huggingface hub.

We should have both a function to load the datasets with the corresponding task:

```python
from distilabel.datasets import load_dataset
dataset: CustomDataset = load_dataset(dataset_name)
```

And push the task to the HF hub automatically
```python
dataset.push_to_hub(dataset_name)
# Include the task as other file in the repository 
```

**Additional context**
This can be tackled after #288

plaguss:
## Description

This PR adds the functionality to push tasks to the hub and download them automatically if available. It will simplify working with `CustomDataset` and keep track of the tasks that come with them out of the box.

An example can be seen [here](https://huggingface.co/datasets/argilla/distilabel-sample-evol-instruct/blob/main/distilabel-task.json).

Closes #290.


src/distilabel/dataset.py
```diff
@@ -210,6 +221,81 @@ def load_from_disk(cls, dataset_path: PathLike, **kwargs: Any):
         ds.task = task
         return ds
 
+    def push_to_hub(
+        self, repo_id: str, *args: Any, push_task: bool = True, **kwargs: Any
```
davidberenstein1957, line 225: Optional type hint?

src/distilabel/dataset.py
```diff
@@ -210,6 +221,81 @@ def load_from_disk(cls, dataset_path: PathLike, **kwargs: Any):
...
+                with tempfile.NamedTemporaryFile(mode="w", delete=False) as f:
+                    f.write(json.dumps(self.task.dump(), indent=2))
+                    f.flush()
+
+                    HfApi().upload_file(
+                        path_or_fileobj=f.name,
+                        path_in_repo=TASK_FILE_NAME,
+                        repo_id=repo_id,
+                        repo_type="dataset",
+                        token=kwargs.get("token"),
```
davidberenstein1957, line 257: if None will this default to the HF token internally?

src/distilabel/dataset.py
```diff
@@ -210,6 +221,81 @@ def load_from_disk(cls, dataset_path: PathLike, **kwargs: Any):
...
+                    )
+            except Exception as e:
+                logger.error(
+                    f"Error while pushing the task to the hub: {e}, task won't be pushed."
```
davidberenstein1957, line 261: I always like to end with the error because otherwise there is this little bit of text at the end of the error.

src/distilabel/dataset.py
```diff
@@ -210,6 +221,81 @@ def load_from_disk(cls, dataset_path: PathLike, **kwargs: Any):
...
+    # download the task
+    try:
+        task_path = hf_hub_download(
+            repo_id=path, filename=TASK_FILE_NAME, repo_type="dataset"
+        )
+        task = load_from_dict(read_json(task_path))
+        cds.task = task
+    except Exception as e:
+        logger.error(
+            f"Error while downloading the task from the hub: {e}, the task won't be loaded."
```
davidberenstein1957, line 295: same here w.r.t. the error.

tests/tasks/test_serialization.py
```diff
@@ -0,0 +1,367 @@
...
+            "module": "distilabel.tasks.preference.ultrajudge",
+            "name": "UltraJudgeTask",
+        },
+        "system_prompt": UltraJudgeTask().system_prompt,
+        "task_description": UltraJudgeTask().task_description,
+        "areas": UltraJudgeTask().areas,
+    }
+
+
+sample_prompetheus_task = PrometheusTask(
```
davidberenstein1957, line 107: why is this defined outside of the fixture?

davidberenstein1957 (commented):
@plaguss looking good! Added some small comments?
 Shouldn't we also test if we can actually push and pull from the hub. 

Also, I think the classes for `Test*TaskSerialization` can be generalized a bit w.r.t. having the same base class and different init args.

src/distilabel/dataset.py
plaguss: The `HfApi().upload_file` function should take care of that on it's own, this is the same implementation we have in argilla to update the dataset card [here](https://github.com/argilla-io/argilla/blob/develop/src/argilla/client/feedback/integrations/huggingface/dataset.py#L232).

src/distilabel/dataset.py
plaguss: Totally

tests/tasks/test_serialization.py
plaguss: No reason at all, updated

PR #300, opened 2024-01-24
Add example of evol instruct dataset

plaguss:
## Description

This PR adds an example of the real definition of the `EvolInstruct` strategy as defined in [WizardLM](https://arxiv.org/abs/2304.12244).
- [x] Review of the documentation.
- [x] Add equal-prompts from wizardlm in the mix.


davidberenstein1957 (commented):
@plaguss awesome 👌

davidberenstein1957 (approved):

sdiazlor:
@plaguss @davidberenstein1957  I added two ways for the evol-instruct to the [tutorial ](https://github.com/argilla-io/distilabel/pull/327), but not sure which one would be better to add to the pipeline, as the `distilabel CustomDataset` can not be returned: if a `Dataset` or a `FeedbackDataset`. I would go for the `Dataset` as it's simpler and quicker for the pipeline, but the `to_argilla` method is then lost.

examples/pipeline-evol-instruct-alpaca.py
```diff
@@ -41,35 +46,142 @@ def prepare(example):
...
+        The Second Prompt: {second_instruction}
+        Your Judgement (Just answer: Equal or Not Equal. No need to explain the reason):"""
+
+    @dataclass
+    class EliminationEqualPrompts(TextGenerationTask):
+        """Task to check for the equality of two instructions following the Appendix G in
+        [WizardLM paper](https://arxiv.org/abs/2304.12244).
+        """
+
+        system_prompt: str = "You are an AI judge in charge of determining the equality of two instructions. "
```
plaguss, line 72:
```suggestion
        system_prompt: str = "You are an AI judge in charge of determining the equality of two instructions."
```

examples/pipeline-evol-instruct-alpaca.py
```diff
@@ -41,35 +46,142 @@ def prepare(example):
...
+            evolved_dataset = evolver_pipe.generate(input_dataset, batch_size=8)
+
+            # Prepare the dataset for the second pipeline with the original
+            # instructions and the evolved ones. If the evolved instruction is None,
+            # we use the original instruction (to make sure it will be removed)
+            def prepare_for_equal_prompts(example):
+                if example["instructions"][0] is None:
+                    return {"input": [example["input"], example["input"]]}
+                else:
+                    return {"input": [example["input"], example["instructions"][0]]}
```
plaguss, line 131: If the function is not changing I think we can put it outside the `for step in ...` loop.

plaguss (commented):
Just a couple of small comments from my side, thank you! 😄

alvarobartt:
What's the status of this @plaguss? Should we close it for the moment and reuse the content on a later stage? Or do you prefer to adapt the current to the latest `EvolInstruct` task on top of the `core-refactor` branch?

PR #312, opened 2024-01-30
fix: typo in clean an existing preference dataset

sdiazlor:
The # typo is now handled and fixes the rendering issue as shown below:

image

### DIFF
--- a/docs/tutorials/clean-preference-dataset-judgelm-gpt.ipynb
+++ b/docs/tutorials/clean-preference-dataset-judgelm-gpt.ipynb
@@ -519,7 +519,7 @@
       "cell_type": "markdown",
       "metadata": {},
       "source": [
-        "## Optional: Post-process the dataset"
+        "### Optional: Post-process the dataset"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


davidberenstein1957:
Hi @sdiazlor , can you decreast the header importance (#=>##) for the headers under "Humand Feedback with Argilla"?

sdiazlor:
@davidberenstein1957 It remains like this:

image

davidberenstein1957:
@sdiazlor that looks better already. I think you might have missed "analyze your cleaned dataset" tough?

davidberenstein1957:
@sdiazlor I was talking about the sidebar. Not sure if you see the difference there?

sdiazlor:
> @sdiazlor that looks better already. I think you might have missed "analyze your cleaned dataset" tough?

I had though of 'analyze your cleaned dataset' as a more important header as it analyzes the final dataset after not only the argilla human feedback but also the distilabel application, that's the reason why, but I can change if you think that it makes more sense

PR #315, opened 2024-01-31
Align the `LLM` subclasses args

### LINKED ISSUE #246: [FEATURE] align argument usage across different llm providers

alvarobartt:
## Description

This PR ensures that the args of `LLM` subclasses are as aligned as possible, with the `model` being the first arg if it has no default value, otherwise the first one will remain `task`. Also the docstrings have been updated accordingly, including complete examples when missing; and the examples have been updated too as some where still using `checkpoint_strategy=True` which is not valid. Also the arg `num_threads` has been included in `LlamaCppLLM` and `vLLM` since it was missing.

Besides that, the arg `openai_api_key` has been renamed to `api_key` to keep consistency with the rest of the `api_key` args provided. Also the `vllm` arg in `vLLM` has been renamed to `model` to keep the consistency with the rest of the args within the `LLM` subclasses.

Closes #246


davidberenstein1957 (commented):
Perhaps we need to update the docs w.r.t. the used arguments and mention we align with the API definitions of specific frameworks for the LLMs.

https://distilabel.argilla.io/latest/technical-reference/llms/

plaguss (commented):
Same as David, I guess a small comment mentioning that for the specific parameters one should go to the API definition used, all of them have already links already.

plaguss (approved):

PR #323, opened 2024-02-02
fix prompt_formatting_fn

### LINKED ISSUE #322: [BUG] prompt_formatting_fn broken
**Describe the bug**
`prompt_formatting_fn` doesn't work at all and introduces subtle errors due to wrong prompting if one doesn't check manually.

**To Reproduce**

Code to reproduce
```python
def format_mistral_prompt(prompt):
    if prompt.system_prompt:
        return f"[INST] {prompt.formatted_prompt}\n\n{prompt.formatted_prompt} [/INST]"
    else:
        return f"[INST] {prompt.formatted_prompt} [/INST]"

llm = TogetherInferenceLLM(
    model="mistralai/Mixtral-8x7B-Instruct-v0.1",
    api_key=os.getenv("TOGETHER_API_KEY", None),
    task=TextGenerationTask(),
    prompt_formatting_fn=format_mistral_prompt,
)

llm.validate_prompts([{"input": "What's a large language model?"}])

> [Prompt(system_prompt='', formatted_prompt="What's a large language model?")]
```

**Expected behaviour**
```python
llm.validate_prompts([{"input": "What's a large language model?"}])

> ["[INST] What's a large language model? [/INST]"]
```

jphme:
fixes #322 .

I don't know if this could break any undocumented legacy use, but `Task` should always return a `Prompt`, so the current implementation just discards `prompt_formatting_fn` silently and the type checks should be unnecessary here?

Code is tested for the problem I encountered in #322 .

### DIFF
--- a/src/distilabel/llm/base.py
+++ b/src/distilabel/llm/base.py
@@ -37,7 +37,6 @@ import multiprocess as mp
 
 from distilabel.llm.utils import LLMOutput
 from distilabel.logger import get_logger
-from distilabel.tasks.prompt import Prompt
 from distilabel.utils.futures import when_all_complete
 
 if TYPE_CHECKING:
@@ -135,17 +134,9 @@ class LLM(ABC):
         prompts = []
         for input in inputs:
             prompt = self.task.generate_prompt(**input)
-            if not isinstance(prompt, Prompt) and self.prompt_formatting_fn is not None:
-                warnings.warn(
-                    "The method `generate_prompt` is not returning a `Prompt` class but a prompt"
-                    f" of `type={type(prompt)}`, meaning that a pre-formatting has already been"
-                    " applied in the `task.generate_prompt` method, so the usage of a `prompt_formatting_fn`"
-                    " is discouraged.",
-                    UserWarning,
-                    stacklevel=2,
-                )
+            if self.prompt_formatting_fn is not None:
                 prompt = self.prompt_formatting_fn(prompt)
-            elif isinstance(prompt, Prompt) and self.prompt_formatting_fn is None:
+            else:
                 if self.prompt_format is not None or default_format is not None:
                     prompt = prompt.format_as(
                         format=self.prompt_format or default_format  # type: ignore


alvarobartt:
Hi here @jphme thanks for the issue and the PR straight away! Our idea with this was mainly for the tasks that were not returning a `Prompt` so that one could define a custom `Task` and then override the prompt-processing function before it gets passed to the `LLM.generate` method. But fair, currently it has no point with the tasks that we provide as part of `distilabel`, so before merging we'll review the current situation of the `prompt_formatting_fn` arg and define whether it's needed or not. Thanks! 🤗

jphme:
> Hi here @jphme thanks for the issue and the PR straight away! Our idea with this was mainly for the tasks that were not returning a `Prompt` so that one could define a custom `Task` and then override the prompt-processing function before it gets passed to the `LLM.generate` method. But fair, currently it has no point with the tasks that we provide as part of `distilabel`, so before merging we'll review the current situation of the `prompt_formatting_fn` arg and define whether it's needed or not. Thanks! 🤗

I figured that this is the reason but in the current implementation `Task` should always return a `Prompt` ([see here](https://github.com/argilla-io/distilabel/blob/8ccf116879fe39041f803a6343b372dbd8635fbe/src/distilabel/tasks/base.py#L73) ) and it also makes sense to always to the prompt formatting stuff on the `LLM` side of things conceptually. 

The current implementation leads to silent errors (prompt format is silently ignored without any warnings or errors) which is the worst choice. And additionally you can`t format the prompts any other way than with the (limited) default templates. My change would make `prompt_formatting_fn` consistent with the default prompt formats (taking precedence if both are provided). 

(But it probably would make sense to improve/adapt the documentation in a few places as well :) )

jphme:
> > Hi here @jphme thanks for the issue and the PR straight away! Our idea with this was mainly for the tasks that were not returning a `Prompt` so that one could define a custom `Task` and then override the prompt-processing function before it gets passed to the `LLM.generate` method. But fair, currently it has no point with the tasks that we provide as part of `distilabel`, so before merging we'll review the current situation of the `prompt_formatting_fn` arg and define whether it's needed or not. Thanks! 🤗
> 
> I figured that this is the reason but in the current implementation `Task` should always return a `Prompt` ([see here](https://github.com/argilla-io/distilabel/blob/8ccf116879fe39041f803a6343b372dbd8635fbe/src/distilabel/tasks/base.py#L73) ) and it also makes sense to always to the prompt formatting stuff on the `LLM` side of things conceptually.
> 
> The current implementation leads to silent errors (prompt format is silently ignored without any warnings or errors) which is imho the worst choice. And additionally you cant format the prompts any other way than with the (limited) default templates. My change would make `prompt_formatting_fn` consistent with the default prompt formats (taking precedence if both are provided).
> 
> (But it probably would make sense to improve/adapt the documentation in a few places as well, happy to help with that if you´ve decided on the preferred way forward :) )

alvarobartt:
> @alvarobartt Any news on the review of `prompt_formatting_fn`?

Hi @jphme, yes, indeed we're currently going through a refactor and we're taking your feedback on this into consideration and the `prompt_formatting_fn` will now be provided to the `LLM`, and it will assume OpenAI as the input format and the the formatting will be applied and converted into whatever format is desired.

We saw some flaws in the current implementation, also thanks to the community issues as this one, and decided to proceed with a refactor that will clear a bit those things and the logic will be more straight forward but at the same time more scalable and flexible. We'll keep you posted, but thanks again for the feedback 🤗

jphme:
> > @alvarobartt Any news on the review of `prompt_formatting_fn`?
> 
> Hi @jphme, yes, indeed we're currently going through a refactor and we're taking your feedback on this into consideration and the `prompt_formatting_fn` will now be provided to the `LLM`, and it will assume OpenAI as the input format and the the formatting will be applied and converted into whatever format is desired.
> 
> We saw some flaws in the current implementation, also thanks to the community issues as this one, and decided to proceed with a refactor that will clear a bit those things and the logic will be more straight forward but at the same time more scalable and flexible. We'll keep you posted, but thanks again for the feedback 🤗

Thanks for letting me know, surely makes sense. When trying to work around this, I encountered another bug (see #350 ) which prevents the use of OpenAI compatible endpoints without a model_list. 

If you´re refactoring currently: Is the `task` parameter at LLM creation really necessary? I know it´s used for tokenization, but imho the API would be alot nicer, if Tasks and LLMs could be created independently of each other.

alvarobartt:
> If you´re refactoring currently: Is the `task` parameter at LLM creation really necessary? I know it´s used for tokenization, but imho the API would be alot nicer, if Tasks and LLMs could be created independently of each other.

Yes, indeed that's the case now, the `LLM` and the `Task` are defined independently and then the `Task` uses the `LLM` to run only the inference, so that the formatting and such is within the `Task` itself.

Long-story-short we went from `LLM has Task` to `Task uses LLM`.

PR #324, opened 2024-02-03
US Spelling and other typo correction on Distilabel tutorials

### LINKED ISSUE #320: [DOCS] Typo "labeller" in ✨ Clean a Preference Dataset with JudgeLMTaskand GPT4-turbo¶
"labeller" -> "labeler". I'll do a general grammatical revision.

ignacioct:
Closes #320.

I changed labeller for labeler and did and overall correction, both with my eyes and Grammarly :)

### DIFF
--- a/docs/tutorials/clean-preference-dataset-judgelm-gpt.ipynb
+++ b/docs/tutorials/clean-preference-dataset-judgelm-gpt.ipynb
@@ -109,7 +109,7 @@
       "source": [
         "### Environment variables\n",
         "\n",
-        "Finally, we will also need to provide a HF_TOKEN and the OPENAI_API_KEY to run the distilabel pipeline."
+        "Finally, we will also need to provide an HF_TOKEN and the OPENAI_API_KEY to run the distilabel pipeline."
       ]
     },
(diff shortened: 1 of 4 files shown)


ignacioct:
> Hi here @ignacioct! FYI we should keep `labeller` for the moment, as the arg within `Pipeline` and `pipeline` is indeed `labeller` and that may imply a breaking change, so we should leave it as `labeller` for the moment IMO. Also note that with @gabrielmbmb upcoming refactor the roles of `generator` and `labeller` may disappear as are right now.

Helo there @alvarobartt ! Shall I then remove all changes regarding labeller, and keep the rest? We then just have to remember to tackle this when that refactor happens, if we wanna stick with US spelling :)

alvarobartt:
Fair @ignacioct! Let's do that for the moment, while taking into consideration the `labeller->labeler` during the refactor of the `Pipeline` 👍🏻

ignacioct:
@alvarobartt just reverted that change. Give it a look and we can merge it :)

plaguss (approved):

Issue #306, opened 2024-01-29
[FEATURE] Integrate reward models like for example `PairRM` to rank LLM responses


plaguss:
**Is your feature request related to a problem? Please describe.**
We could integrate [`PairRM`](https://huggingface.co/llm-blender/PairRM) from [LLMBlender](https://github.com/yuchenlin/LLM-Blender) to rank responses from the LLMs.

**Describe the solution you'd like**
I would like to have a model to rank LLM responses using `PairRM` integrated in `distilabel`.

**Describe alternatives you've considered**
.

**Additional context**
It may need a new type of `Task` to deal with `ranking` questions instead of `rating` ones (take a look at the argilla integration).

plaguss:
Closed in favor of #433

PR #302, opened 2024-01-25
Add `EvolComplexityScorer` and `EvolQualityScorer` tasks from Deita

plaguss:
## Description

This PR adds new tasks `EvolComplexityScorer` and `EvolQualityScorer` from the Deita paper.

Example of use (**Complexity**):

```python
from datasets import load_dataset

ds = load_dataset("argilla/distilabel-intel-orca-dpo-pairs", split="train")
subset = ds.select(range(2))
```

From an LLM:

```python
from distilabel.tasks.preference.evol_scorer import EvolComplexityScorerTask
from distilabel.llm import OpenAILLM
from distilabel.pipeline import Pipeline

llm = OpenAILLM(
    EvolComplexityScorerTask(),
    openai_api_key=os.getenv("OPENAI_API_KEY", None),
    temperature=0.3
)
res = llm.generate([{"generations": subset[0]["generations"]}])
# >>> print(res[0][0]["raw_output"])
# [1] Score: 1
# [2] Score: 2
# >>> print(res[0][0]["parsed_output"])
# {'ranks': [1, 2]}
```

From a Pipeline

```python
>>> pipe = Pipeline(labeller=llm)
>>> new_ds = pipe.generate(subset)
>>> new_ds.select_columns(["generations", "ranks"])[:]
{'generations': [['[\n  ["AFC Ajax (amateurs)", "has ground", "Sportpark De Toekomst"],\n  ["Ajax Youth Academy", "plays at", "Sportpark De Toekomst"]\n]',
   " Sure, I'd be happy to help! Here are the RDF triplets for the input sentence:\n\n[AFC Ajax (amateurs), hasGround, Sportpark De Toekomst]\n[Ajax Youth Academy, playsAt, Sportpark De Toekomst]\n\nExplanation:\n\n* AFC Ajax (amateurs) is the subject of the first triplet, and hasGround is the predicate that describes the relationship between AFC Ajax (amateurs) and Sportpark De Toekomst.\n* Ajax Youth Academy is the subject of the second triplet, and playsAt is the predicate that describes the relationship between Ajax Youth Academy and Sportpark De Toekomst.\n\nNote that there may be other possible RDF triplets that could be derived from the input sentence, but the above triplets capture the main relationships present in the sentence."],
  ['Midsummer House is a moderately priced Chinese restaurant with a 3/5 customer rating, located near All Bar One.',
[...]

PR #329, opened 2024-02-05
Add examples for the deita paper tasks

plaguss:
## Description

This PR includes additional examples and adds some notes for the Deita tasks.

### DIFF
--- a/docs/snippets/technical-reference/llm/mistralai_generate.py
+++ b/docs/snippets/technical-reference/llm/mistralai_generate.py
@@ -8,7 +8,7 @@ mistralai_llm = MistralAILLM(
     task=TextGenerationTask(),
     api_key=os.environ.get("MISTRALAI_API_KEY"),
 )
-result = mistralai_llm.generate([{"input": "What is Anyscale?"}])
+result = mistralai_llm.generate([{"input": "What's the best french cheese?"}])
 # >>> print(result[0][0]["parsed_output"]["generations"])
 # I'd be happy to help answer your question, but it's important to note
 # that the "best" French cheese can be subjective as it depends on personal
--- a/docs/snippets/technical-reference/tasks/complexity_scorer_example.py
+++ b/docs/snippets/technical-reference/tasks/complexity_scorer_example.py
@@ -1,0 +1,53 @@
+import os
+from datasets import Dataset
+
+from distilabel.tasks import ComplexityScorerTask
+from distilabel.llm import OpenAILLM
+from distilabel.pipeline import Pipeline
+
+
+# Create a sample dataset (this one is inspired from the distilabel-sample-evol-complexity dataset)
+sample_evol_complexity = Dataset.from_dict(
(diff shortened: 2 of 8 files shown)


davidberenstein1957 (commented):
Hi @plaguss, I realized we return the rank/scores here but I'm not sure if this is handled intuitively because normally we work with ratings where a higher rating corresponds to a higher score but in this case, it is the other way around. Right? And this might not be aligned with the `RankingQuestion` in Argilla itself.

plaguss:
Hi @davidberenstein1957, we don't really follow the paper strictly, we decided to interpret the scores as if those were `Ratings`, so we can reuse all the behaviour from the `PreferenceTask` without adding extra complexity for these tasks. This way, they are more reusable and we don't need to add a specific type of `Task` which takes into account the rankings. We should tackle this when we start integrating `PairRM` and such (this issue should take those into account: https://github.com/argilla-io/distilabel/issues/306). In this case the higher ratings are also higher scores too, higher complexity/quality corresponds to a better response.

The discussion from the PR can be seen [here](https://github.com/argilla-io/distilabel/pull/302).

davidberenstein1957:
@plaguss I agree but now it seems like we attribute ratings to values that are originally rankings so they should be reversed to allow for choosing on of the preferred options. Correct?

rank 1, 2 => rating 2, 1 because a lower rating should translate in a higher rank?

plaguss:
> @plaguss I agree but now it seems like we attribute ratings to values that are originally rankings so they should be reversed to allow for choosing on of the preferred options. Correct?
> 
> rank 1, 2 => rating 2, 1 because a lower rating should translate in a higher rank?
> 
> How is this handled in the to_argilla for example?

I understand, but for example, the `QualityScorer` uses the following prompt:
```python
from distilabel.tasks import QualityScorerTask

task = QualityScorerTask()
print(task.generate_prompt("What are the first 5 Fibonacci numbers?", ["0 1 1 2 3", "0 1 1 2 3"]).formatted_prompt)
```
```txt
Rank the following responses provided by different AI assistants to the user’s question
according to the quality of their response. Score each response from 1 to 2, with 3
reserved for responses that are already very well written and cannot be improved further.
Your evaluation should consider factors such as helpfulness, relevance, accuracy, depth,
creativity, and level of detail of the response.
Use the following format:
[Response 1] Score:
[Response 2] Score:
...
#Question#: What are the first 5 Fibonacci numbers?
#Response List#:

[Response 1] 0 1 1 2 3
[Response 2] 0 1 1 2 3
```

Maybe we can discuss whether the prompts really make sense in our case and maybe rewrite them to more more "inspired" from the deita paper, but just to represent proper ratings, explaining if lower/higher is worse/better

davidberenstein1957:
@plaguss Yes it makes sense. I thought we needed to rank the responses within the prompt and that they were not directly asked to score. Ignore what I've commented above. 😅

PR #337, opened 2024-02-08
docs: update tutorials avoid argilla installation error

sdiazlor:
* Moved the `pip` installation of argilla to the top to avoid importing errors later.
* Some unneeded output removal.
* Some updates of the sections on argilla


davidberenstein1957:
@sdiazlor, loooking good. I think we can just use one `pip install [*, argilla]` call. I think I saw `pip install [openai]` somewhere

davidberenstein1957 (approved):

PR #339, opened 2024-02-08
Clalrify number of generations produced when using LLMPool in docs

davanstrien:
IUC the LLMPool will generate the target number of generations specified when calling `generate` (either using all LLMs or a random selection) i.e. https://github.com/argilla-io/distilabel/blob/a701e838c9aa7809d6674a815a9376c64e451590/src/distilabel/llm/base.py#L704


gabrielmbmb (approved):

gabrielmbmb:
Thanks @davanstrien!

PR #344, opened 2024-02-14
Refactor _build_dataset piece for speed

plaguss:
## Description

This PR does a small refactor in `Pipeline._build_dataset` to build the final dataset using `Dataset.from_pandas` instead of iteratively calling `Dataset.add_item`, as that made the process prohibitively slow for bigger datasets (5K rows and onwards started to be a problem).


alvarobartt (approved):
💯

PR #349, opened 2024-02-20
Add serialization for the new components

plaguss:
## Description

This PR adds methods to serialize to the `Step`, `DAG` and `BasePipeline` classes.

It includes a new `test/integration`, for the moment just with a simple pipeline without LLMs to check the behaviour with something more complex than the unit tests, but we can remove those.


src/distilabel/utils/serialization_v2.py
```diff
@@ -0,0 +1,231 @@
...
+        but for the DAG class, we networkx to obtain the information.
+
+        Args:
+            obj (Any): The object to be serialized.
+
+        Returns:
+            Dict[str, Any]: Data to be dumped for the object.
+        """
+        # NOTE(plaguss): This should be reimplemented for some cases, so maybe we have to enforce it via ABC?.
+        # It will work out of the box for pydantic's BaseModels.
```
alvarobartt, line 129: FYI all the steps will be `pydantic.BaseModel` as those will need to inherit from `Step`, so we're safe assuming everything's a `pydantic.BaseModel` AFAIK cc @gabrielmbmb

tests/pipeline/test_base.py
```diff
@@ -11,7 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import TYPE_CHECKING
+import tempfile
```
alvarobartt, line 14:
```suggestion

import tempfile
```

alvarobartt (commented):
LGTM, and AFAIK you didn't touch any of the previously existing files which is nice because we may do a cleanup so that we remove the old stuff from `core-refactor` to avoid having code that's not necessary anymore!

src/distilabel/utils/serialization_v2.py
plaguss: The comment was related to the classes we will serialize other than the `Step`, the `DAG` for example has to update this method. But for the general case of the `Step` that works, so I guess I wanted that method to let as some freedom to tweak the process without writing to much extra code.

src/distilabel/pipeline/_dag.py
```diff
@@ -28,11 +28,13 @@
 
 import networkx as nx
 
+from distilabel.utils.serialization_v2 import _get_class, _Serializable
```
alvarobartt, line 31: Maybe you can move this to `src/distilabel/pipeline/serialization.py` temporarily instead, so as to avoid having conflicts with the previous files (we will remove them all)

Also the new stuff is only under `src/distilabel/pipeline` so that may be clearer, and if you're OK with that I can fill a PR afterwards removing the previous/outdated files 👍🏻

src/distilabel/pipeline/_dag.py
```diff
@@ -307,3 +309,52 @@ def _validate_step_process_runtime_parameters(self, step: "Step") -> None:
                     f"Step '{step.name}' is missing required runtime parameter '{param_name}'."
                     " Please, provide a value for it when calling `Pipeline.run`"
                 )
+
+    def _model_dump(self, obj: Any, **kwargs: Any) -> Dict[str, Any]:
+        """Dumps the content of the DAG to a dict.
+
+        References:
+        https://networkx.org/documentation/stable/reference/readwrite/generated/networkx.readwrite.json_graph.adjacency_data.html#networkx.readwrite.json_graph.adjacency_data
```
alvarobartt, line 317: That's the same formatting we were applying before just so that's more readable in the docs

```suggestion
        References:
        * [`adjacency_data` - NetworkX Documentation](https://networkx.org/documentation/stable/reference/readwrite/generated/networkx.readwrite.json_graph.adjacency_data.html#networkx.readwrite.json_graph.adjacency_data)
```

src/distilabel/utils/serialization_v2.py
```diff
@@ -0,0 +1,256 @@
...
+
+import yaml
+from pydantic import BaseModel
+
+T = TypeVar("T")
+
+DISTILABEL_FILENAME = "distilabel-file.json"
+
+
+SaveFormat = Literal["json", "yaml"]
```
alvarobartt, line 29: `SaveFormat -> SaveFormats`

```suggestion
SaveFormats = Literal["json", "yaml"]
```

src/distilabel/utils/serialization_v2.py
```diff
@@ -0,0 +1,256 @@
...
+    mod = importlib.import_module(module)
+    return getattr(mod, name)
+
+
+def load_from_dict(class_: Dict[str, Any]) -> Generic[T]:
+    """Reads a template (a class serialized) and returns the instance
+    contained.
+
+    Args:
+        template (Dict[str, Any]): Dict containing the template, the dict serialized.
```
alvarobartt, line 42: Should `class_` be renamed to `template` or the other way around?

```suggestion
def load_from_dict(class_: Dict[str, Any]) -> Generic[T]:
    """Reads a template (a class serialized) and returns the instance
    contained.

    Args:
        class_ (Dict[str, Any]): Dict containing the template, the dict serialized.
```

tests/integration/test_pipe_llms.py
```diff
@@ -0,0 +1,125 @@
...
+    def format_output(self, output: str) -> Dict[str, Any]:
+        return {self.outputs[0]: output}
+
+    def process(self, inputs: StepInput) -> Generator[List[Dict[str, Any]], None, None]:
+        for input in inputs:
+            formatted_input = self.format_input(input)
+            output = self.llm.generate(formatted_input)
+            formatted_output = self.format_output(output)
+            input.update(formatted_output)
+        yield inputs
```
alvarobartt, line 94: This is outdated, feel free to use `distilabel.pipeline.step.task.base import Task` or `distilabel.pipeline.step.task.generation import TextGeneration`

tests/integration/test_pipe_llms.py
```diff
@@ -0,0 +1,125 @@
...
+            name="generate_response",
+            llm=OpenAILLM(api_key="sk-***"),
+            output_mapping={"generation": "output"},
+        )
+        rename_columns.connect(generate_response)
+
+        generate_response_mini = Task(
+            name="generate_response_mini",
+            llm=TransformersLLM(model="TinyLlama/TinyLlama-1.1B-Chat-v1.0"),
+            output_mapping={"generation": "output"},
```
alvarobartt, line 113: If you change the code above, note that the `output_mapping` is no longer part of `Task` neither `TextGeneration`

tests/pipeline/llm/test_serialization.py
```diff
@@ -0,0 +1,128 @@
...
+    def format_output(self, output: str) -> Dict[str, Any]:
+        return {self.outputs[0]: output}
+
+    def process(self, inputs: StepInput) -> Generator[List[Dict[str, Any]], None, None]:
+        for input in inputs:
+            formatted_input = self.format_input(input)
+            output = self.llm.generate(formatted_input)
+            formatted_output = self.format_output(output)
+            input.update(formatted_output)
+        yield inputs
```
alvarobartt, line 92: Same here!

alvarobartt (changes requested):
LGTM, but I've left some comments to tackle before merging! Great work 💪🏻

src/distilabel/pipeline/_dag.py
plaguss: Totally. The `src/distilabel/pipeline/serialization.py` is done

alvarobartt (approved):

PR #361, opened 2024-02-26
Migrate vllm to the new API

plaguss:
## Description

This PR adds support for vllm in the new pipeline.

```python
llm = vLLM(
    model="microsoft/phi-2",
    model_kwargs={
        "tensor_parallel_size": 1,
        "trust_remote_code": True
    }
)
llm.load()
input_ =[
  {
    "content": "Every day, a tree drops 7 leaves. How many leaves would it drop in a month of February in a non-leap year? Include your logic.",
    "role": "user"
  },
  {
    "content": "Here's the logic behind this:\n\n1. We know that February has 28 days in a non-leap year.\n2. If the tree drops 7 leaves every day, then over the course of February, it would drop:\n Leaves dropped in February = Leaves per day * Days in February\n = 7 leaves * 28 days\n = 196 leaves\n\nSo, the tree would drop 196 leaves in February in a non-leap year.",
    "role": "assistant"
  }
]
llm2.generate(input_)
```

### DIFF
--- a/src/distilabel/pipeline/llm/vllm.py
+++ b/src/distilabel/pipeline/llm/vllm.py
@@ -1,0 +1,66 @@
+# Copyright 2023-present, Argilla, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict, Optional
+
+from vllm import LLM as _vLLM
+from vllm import SamplingParams
+
+from distilabel.pipeline.llm.base import LLM
+from distilabel.pipeline.step.task.typing import ChatType
+
+
+class vLLM(LLM):
+    model: str
+    model_kwargs: Optional[Dict[str, Any]] = {}
+    chat_format: Optional[str] = None
+
+    def load(self) -> None:
+        self._model = _vLLM(self.model, **self.model_kwargs)  # type: ignore
+        self._tokenizer = self._model.get_tokenizer()
+        self._values["model_name"] = self.model
+
+    def format_input(self, input: ChatType) -> ChatType:
+        return self._tokenizer.apply_chat_template(  # type: ignore
+            input,
+            tokenize=False,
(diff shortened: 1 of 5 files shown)


src/distilabel/pipeline/llm/vllm.py
```diff
@@ -0,0 +1,66 @@
...
+    model: str
+    model_kwargs: Optional[Dict[str, Any]] = {}
+    chat_format: Optional[str] = None
+
+    def load(self) -> None:
+        self._model = _vLLM(self.model, **self.model_kwargs)  # type: ignore
+        self._tokenizer = self._model.get_tokenizer()
+        self._values["model_name"] = self.model
+
+    def format_input(self, input: ChatType) -> ChatType:
```
alvarobartt, line 34: Maybe can temporarily rename this method to something else, as the `Task` method is also called `format_input`.

alvarobartt, line 34:
```suggestion
    def format_input(self, input: ChatType) -> str:
```

src/distilabel/pipeline/llm/vllm.py
```diff
@@ -0,0 +1,66 @@
...
+
+class vLLM(LLM):
+    model: str
+    model_kwargs: Optional[Dict[str, Any]] = {}
+    chat_format: Optional[str] = None
+
+    def load(self) -> None:
+        self._model = _vLLM(self.model, **self.model_kwargs)  # type: ignore
+        self._tokenizer = self._model.get_tokenizer()
+        self._values["model_name"] = self.model
```
alvarobartt, line 32: Poll time! What do you prefer here, to re-use the `_values["model_name"]` or to define here the `property` instead?

i.e. the current over the following? or the other way around?

```python
@property
def model_name(self) -> str:
    return self.model
```

plaguss: I personally prefer de property `model_name` because I find it more straightforward, but not a strong opinion against the `self._values`

src/distilabel/pipeline/llm/vllm.py
plaguss: The other `LLMs` use the same method, how should we rename `format_input`?

src/distilabel/pipeline/llm/vllm.py
alvarobartt: Cool, then feel free to update the rest of the `LLM`s so that we use the property instead 🤗

src/distilabel/pipeline/llm/vllm.py
alvarobartt: Not sure, just mentioned to avoid duplication but I don't have anything in mind 😩

plaguss: `def prepare_input(self, input: ChatType) -> Any:` looks good to me, we can update it later 😄

PR #366, opened 2024-02-29
Update `_BatchManager` to work with `GlobalStep`s and `input_batch_size` per step

gabrielmbmb:
## Description

This PR updates the `_BatchManager` to handle the creation of a "super" batch gathering all the data from the previous batches for the `GlobalStep`s. In addition, it adds a new attribute to the `Step` class called `input_batch_size`. This argument can be used to define the desired number of rows that the step will receive per batch. Having that said, the `_BatchManager` has also been updated to store the data of the output batches, instead of the batches itself. This allows to build batches of the desired size for each step.

### DIFF
--- a/src/distilabel/utils/serialization_v2.py
+++ b/src/distilabel/utils/serialization_v2.py
@@ -29,7 +29,7 @@ DISTILABEL_FILENAME = "distilabel-file.json"
 SaveFormats = Literal["json", "yaml"]
 
 
-def _get_class(module: str = None, name: str = None) -> Type:
+def _get_class(module: str, name: str) -> Type:
     mod = importlib.import_module(module)
     return getattr(mod, name)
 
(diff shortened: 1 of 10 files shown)


src/distilabel/utils/serialization_v2.py
```diff
@@ -29,7 +29,7 @@
 SaveFormats = Literal["json", "yaml"]
 
 
-def _get_class(module: str = None, name: str = None) -> Type:
+def _get_class(module: str, name: str) -> Type:
```
plaguss, line 32: The default args are placed so [here](https://github.com/argilla-io/distilabel/pull/366/files#diff-a566a0dfe56d6cf771f7e75367798fb6e01641133dce994665d5e256699a21a0R355) we can pass the data using **kwargs. But also, I see I have to update the code to use `distilabel/pipeline/serialization` and remove this file. Will do that in a different PR.

plaguss (commented):
Just a comment irrelevant to the nice PR

src/distilabel/utils/serialization_v2.py
gabrielmbmb: oops, my bad!

PR #371, opened 2024-03-01
Move batching from `Task` to `LLM`, fix `vLLM.generate` and add `DISTILABEL_LOG_LEVEL`

alvarobartt:
## Description

This PR moves the batching from the `Task` to the `LLM` so that the `LLM` handles the batches in the best way possible rather than via a simple for-loop, since there are some LLM engines that have mechanisms to handle batches in a more efficient way. Also the `prepare_input` abstract method has been removed from `LLM` and is no longer needed, unless a specific LLM implementation requires it.

Also this PR fixes `vLLM.generate` and stops propagating the unsolicited `inputs` through the `Pipeline`, so that only the ones solicited via the `property` are kept.

Besides that, this PR also adds the `DISTILABEL_LOG_LEVEL` environment variable to control the log level of `distilabel`, which defaults to `INFO`.

## Example

Find a full example at https://huggingface.co/datasets/alvarobartt/instruction-dataset-mistral-7b-instruct-v0.2/blob/main/example.py

### DIFF
--- a/src/distilabel/pipeline/llm/base.py
+++ b/src/distilabel/pipeline/llm/base.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, Dict, List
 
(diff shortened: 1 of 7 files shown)


plaguss (approved):
🚀

PR #381, opened 2024-03-05
Add `AsyncOpenAI` and update `OpenAILLM` accordingly

alvarobartt:
## Description

This PR adds `AsyncLLM` adding a new abstract method to implement in its subclasses named `agenerate` that expects a single input instead of a list of inputs in order to be able to parallelize the requests to the async-compatible `LLM` subclasses i.e. `OpenAILLM`. Also the `generate` method is implemented within `AsyncLLM` so that the logic is handled already, meaning that only the `async agenerate` needs to be implemented for each subclass.

Besides that, this PR updates the `OpenAILLM` to use `AsyncOpenAI` instead of the synchronous `OpenAI` client.

Finally, some docstrings have been included to both `AsyncLLM` and `OpenAILLM`, as well as unit tests for `OpenAILLM`.

Closes [#380](https://github.com/argilla-io/distilabel/issues/380)

### Update

Additionally, this PR also solves a bug with the `field_validator` of `OpenAILLM.api_key`, read more about it at https://docs.pydantic.dev/latest/concepts/validators/#validation-of-default-values. And this PR also fixes the `GlobalStep` since it was inheriting from `_Step` instead of `Step`, so it was breaking.


src/distilabel/llm/base.py
```diff
@@ -0,0 +1,78 @@
...
+        pass
+
+    @abstractmethod
+    def generate(
+        self, inputs: List["ChatType"], *args: Any, **kwargs: Any
+    ) -> List[str]:
+        pass
+
+
+class AsyncLLM(LLM):
```
gabrielmbmb, line 49: noice

gabrielmbmb (approved):

PR #405, opened 2024-03-11
Add `OllamaLLM`

davidberenstein1957:


src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Args:
+        model: the model name to use for the LLM e.g. "notus", etc.
+        host: the host to use for the LLM e.g. "https://ollama.com".
+        max_retries: the maximum number of retries for the LLM.
+        timeout: the timeout for the LLM.
+
+    """
+
+    model: str = "notus"
```
alvarobartt, line 39: IMO we shouldn't set `model` with a default

```suggestion
    model: str
```

src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,75 @@
...
+    host: Union[str, None] = None
+    timeout: int = 120
+    follow_redirects: bool = True
+
+    _aclient: Optional["AsyncClient"] = PrivateAttr(...)
+
+    def load(self) -> None:
+        """Loads the `AsyncClient` and `Client` for Ollama."""
```
alvarobartt, line 47:
```suggestion
        """Loads the `AsyncClient` to use Ollama async API."""
```

src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,75 @@
...
+        self._aclient = AsyncClient(
```
alvarobartt, line 48: Can the Ollama APIs be protected with an `api_key`? If so, we should include it, I'll create a PR today to include that for Mistral and OpenAI so feel free to reuse (also to be able to provide the API keys in runtime).

src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,75 @@
...
+        return self.model
+
+    async def agenerate(
+        self,
+        input: "ChatType",
+        stream: bool = False,
+        format: Literal["", "json"] = "",
+        options: Union["Options", None] = None,
+    ) -> str:
+        """Generates a response asynchronously, using the Ollama Async API definition."""
```
alvarobartt, line 66: Here the docstring should contain the description for each of the arguments above i.e. the whole docstring with args, raises and returns.

src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @property
+    def model_name(self) -> str:
+        """Returns the model name used for the LLM."""
+        return self.model
+
+    async def agenerate(
+        self,
+        input: "ChatType",
+        stream: bool = False,
```
alvarobartt, line 62: We won't benefit from streaming in any case, so we should disable by default and exclude it from the `args` IMO

```suggestion
```

src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,75 @@
...
+        format: Literal["", "json"] = "",
+        options: Union["Options", None] = None,
+    ) -> str:
+        """Generates a response asynchronously, using the Ollama Async API definition."""
+        completion = await self._aclient.chat(  # type: ignore
+            model=self.model,
+            messages=input,
+            stream=stream,
```
alvarobartt, line 70: Related to the `stream` arg exclusion mentioned above.

```suggestion
            stream=False,
```

src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def agenerate(
+        self,
+        input: "ChatType",
+        stream: bool = False,
+        format: Literal["", "json"] = "",
+        options: Union["Options", None] = None,
+    ) -> str:
+        """Generates a response asynchronously, using the Ollama Async API definition."""
+        completion = await self._aclient.chat(  # type: ignore
```
alvarobartt, line 67: Does the Ollama client allow something like `n` for the OpenAI client to control the number of generations per input? Also the current args look limited to me, is there any way we can include more sampling/generation args?

tests/unit/llm/test_ollama.py
```diff
@@ -0,0 +1,95 @@
...
+
+try:
+    from distilabel.llm.ollama import OllamalLLM
+except ImportError:
+    pass
+
+
+@patch("ollama.AsyncClient")
+class TestOllamaLLM:
+    def test_ollama_llm(self, mock_ollama: MagicMock) -> None:
```
alvarobartt, line 28:
```suggestion
    def test_ollama_llm(self) -> None:
```

src/distilabel/llm/ollama.py
davidberenstein1957: It cannot

src/distilabel/llm/ollama.py
davidberenstein1957: The current args are limited because they are all captured in the `options` which is the normal behavior for `ollama`. Additionally, a lot of these variables are defined in the model card.

davidberenstein1957: And it does not allow for `n` per input.

src/distilabel/llm/ollama.py
```diff
@@ -0,0 +1,99 @@
...
+            num_generations: the number of generations to produce. Defaults to `1`.
+            format: the format to use for the generation. Defaults to "".
+            options: the options to use for the generation. Defaults to `None`.
+            keep_alive: whether to keep the connection alive. Defaults to `None`.
+
+        Returns:
+            A list of strings as completion for the given input.
+        """
+        generations = []
+        for _ in range(num_generations):
```
gabrielmbmb, line 88: Instead of doing this for-loop, I would override the `generate` method like here: https://github.com/argilla-io/distilabel/blob/901fe7ddaf1a284521c37759a2d052a8f5083963/src/distilabel/llm/mistral.py#L141-L164

gabrielmbmb (changes requested):
LGTM! If we can remove the for-loop before merging that would be nice

PR #407, opened 2024-03-11
Add `EvolInstruct` and `EvolInstructGenerator` tasks

### LINKED ISSUE #408: Add `EvolInstruct` faithful reproduction of `WizardLM`

alvarobartt:
## Description

This PR adds both the `EvolInstruct` and `EvolInstructGenerator` tasks ported from https://github.com/h2oai/h2o-wizardlm/blob/main/wizardlm.py with some slight modifications to suit our needs, but respecting the evolutionary approach.

Besides that the `AsyncLLM` has been fixed so as to use `asyncio` event loops instead of `asyncio.run` as it was raising some errors when called within a loop, so now the `AsyncLLM` implementation is more robust. Also the `GeneratorTask` has been included for tasks like `EvolInstructGenerator` i.e. generating data without seed data as input.

Closes #408 

## Example

```python
import time

from distilabel.llm.mistral import MistralLLM
from distilabel.steps.globals.huggingface import PushToHub
from distilabel.steps.task.evol_instruct.generator import EvolInstructGenerator

if __name__ == "__main__":
    start_time = time.time()

    with Pipeline() as pipeline:
        evol_instruct = EvolInstructGenerator(
            name="evol_instruct",
            llm=OpenAILLM(
                model="gpt-4",
                api_key="sk-***",  # type: ignore
            ),
            num_instructions=10,
            generate_answers=False,
        )

        push_to_hub = PushToHub(name="push_to_hub")  # type: ignore
        evol_instruct.connect(push_to_hub)

    pipeline.run(
        parameters={
            "evol_instruct": {
                "generation_kwargs": {
                    "max_new_tokens": 512,
                    "temperature": 0.7,
                },
            },
            "push_to_hub": {
                "repo_id": "alvarobartt/evol-instruct",
                "split": "train",
                "private": False,
                "token": "hf_***",
            },
        }
    )

```

## Reference

https://github.com/h2oai/h2o-wizardlm/blob/main/wizardlm.py


davidberenstein1957 (commented):
Hi @alvarobartt, I am getting an error.

```
Traceback (most recent call last):
  File "/.pyenv/versions/3.10.6/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/.pyenv/versions/3.10.6/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/.pyenv/versions/3.10.6/lib/python3.10/multiprocessing/pool.py", line 595, in _handle_results
    cache[job]._set(i, obj)
  File "/.pyenv/versions/3.10.6/lib/python3.10/multiprocessing/pool.py", line 781, in _set
    self._error_callback(self._value)
  File "/distilabel/src/distilabel/pipeline/local.py", line 195, in _error_callback
    if e.is_load_error:
AttributeError: 'MaybeEncodingError' object has no attribute 'is_load_error'
```
This might have been introduced by the changes to `llm/base.py` @gabrielmbmb is having a look.

src/distilabel/steps/task/evol_instruct/base.py
```diff
@@ -0,0 +1,213 @@
...
+    def format_input(self, input: str) -> ChatType:  # type: ignore
+        """The input is formatted as a `ChatType` assuming that the instruction
+        is the first interaction from the user within a conversation. And the
+        `system_prompt` is added as the first message if it exists."""
+        return [{"role": "user", "content": input}]
+
+    @property
+    def outputs(self) -> List[str]:
+        """The output for the task is the `generation` and the `model_name`."""
+        # TODO: having to define a `model_name` column every time as the `Task.outputs` is not ideal,
```
davidberenstein1957, line 93: might be interesting for @gabrielmbmb

src/distilabel/steps/task/evol_instruct/generator.py
```diff
@@ -0,0 +1,244 @@
...
+class EvolInstructGenerator(GeneratorTask):
+    """WizardLM: Empowering Large Language Models to Follow Complex Instructions
+
+    Reference:
+        - https://arxiv.org/abs/2304.12244
+        - https://github.com/h2oai/h2o-wizardlm
+    """
+
+    num_instructions: int
+    generate_answers: bool = False
```
davidberenstein1957, line 53: why don't we offer to store evolutions here?

src/distilabel/steps/task/evol_instruct/generator.py
```diff
@@ -0,0 +1,244 @@
...
+    mutation_templates: EnumType = Field(default=GenerationMutationTemplates)
+
+    min_length: RuntimeParameter[int] = Field(default=256)
+    max_length: RuntimeParameter[int] = Field(default=1024)
+
+    _seed_texts: Optional[List[str]] = PrivateAttr(default_factory=list)
```
davidberenstein1957, line 59: It seems this is a private attribute but within WizardLM they give the options to use this to lad a file or pass a list of seed values yourself.

davidberenstein1957:
@alvarobartt Also, I seem to be missing the elimination but you wanted to skip this because general state of LLMs have improved and you feel we don't need it anymore, correct?
image

davidberenstein1957:
@alvarobartt, I am also missing the system prompt for the LLM but we don't want to implement that either? https://github.com/argilla-io/distilabel/blob/main/src/distilabel/tasks/_templates/evol-instruct.jinja2

alvarobartt:
> @alvarobartt, I am also missing the system prompt they defined in the paper but we don't want to implement that either? https://github.com/argilla-io/distilabel/blob/main/src/distilabel/tasks/_templates/evol-instruct.jinja2

Fair, we can include it, but I'm afraid that the misalignments between their official implementation and the paper don't have a clear reference, should we take more inspiration from the paper instead? My experience so far is that using GPT-4 for generating and evolving instructions works well so far, but we can try to compare with and without `system_prompt`. Also does that apply to both the `instruction` and the `answer`, or only the `instruction`? I guess the first one, but LMK

> @alvarobartt Also, I seem to be missing the elimination but you wanted to skip this because general state of LLMs have improved and you feel we don't need it anymore, correct? image

Right, I was using `gpt-4`, `mistral-medium` and `mistral-large`, and didn't run into any of those issues to tackle during post-processing, but maybe we can include it just in case, but IMO those were not needed for this small proof of concept. Anyway, all these details will show whenever we implement DEITA using this building blocks to see whether something's off or not, but I'm afraid that some things they added on top of the post-processing where suited for their models, while bigger and more recent ones seem to preduce nice results with minimal post-processing.

src/distilabel/steps/task/evol_instruct/generator.py
alvarobartt: Sure, we could add that flag too!

src/distilabel/steps/task/evol_instruct/generator.py
```diff
@@ -0,0 +1,244 @@
...
+    Reference:
+        - https://arxiv.org/abs/2304.12244
+        - https://github.com/h2oai/h2o-wizardlm
+    """
+
+    num_instructions: int
+    generate_answers: bool = False
+    mutation_templates: EnumType = Field(default=GenerationMutationTemplates)
+
+    min_length: RuntimeParameter[int] = Field(default=256)
```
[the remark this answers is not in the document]
alvarobartt: Oops probably a debugging left over (to end the generation earlier as the threshold was bypassed before)

src/distilabel/steps/task/evol_instruct/generator.py
```diff
@@ -0,0 +1,244 @@
...
+
+class EvolInstructGenerator(GeneratorTask):
+    """WizardLM: Empowering Large Language Models to Follow Complex Instructions
+
+    Reference:
+        - https://arxiv.org/abs/2304.12244
+        - https://github.com/h2oai/h2o-wizardlm
+    """
+
+    num_instructions: int
```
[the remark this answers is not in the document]
alvarobartt: Didn't experience that myself either, but sure! Also note that the `seed_texts` will always contain N instructions to evolve, where N is equal to the `num_instructions` to generate, which IMO is suboptimal in some scenarios, but overall alleviates the issue with infinite non-evolvable instructions. Anyway, worth having a look, is that somewhere in the paper and/or their implementation? Maybe I just missed it!

alvarobartt:
Also I've been exploring a bit on GitHub and, is this a more faithful reproduction of EvolInstruct (pre-WizardLM)? https://github.com/nlpxucan/WizardLM/tree/main/Evol_Instruct cc @davidberenstein1957

davidberenstein1957:
@alvarobartt I would say so, yes. This is also the way we eventually implemented it initially.

src/distilabel/steps/task/evol_instruct/generator.py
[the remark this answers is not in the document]
davidberenstein1957: No it is not but normally I would expect it to run x generations, where x is an input argument, like with the non-generator. In my example when I ran the pipeline it took 4 generations to evolve an evolution from scratch, so I thought it might have a risk of racking up un-expected costs/run time for some people.

src/distilabel/steps/task/evol_instruct/base.py
```diff
@@ -0,0 +1,213 @@
...
+                    for evolved_instruction in evolved_instructions
+                ]
+
+            self._logger.info(
+                f"🔄 Ran iteration {iter_no} evolving {len(instructions)} instructions!"
+            )
+
+        if self.store_evolutions:
+            # Remove the input instruction from the `evolved_instructions` list
+            instructions = [instruction[1:] for instruction in instructions]
```
gabrielmbmb, line 185: I think for downstream steps using the output of this one, it would be easier to also include the original instruction.

src/distilabel/steps/task/evol_instruct/base.py
```diff
@@ -0,0 +1,213 @@
...
+            _output[self.output_mappings["evolved_instructions"]] = instructions
+
+        if self.generate_answers and answer is not None:
+            _output[self.output_mappings["answer"]] = answer
+
+        _output[self.output_mappings["model_name"]] = self.llm.model_name
+        return _output
+
+    @override
+    def process(self, inputs: "StepInput") -> "StepOutput":  # type: ignore
```
gabrielmbmb, line 127: I think we could split this method in two major methods, that then are used in `process`: `_evolve_instructions` and `_generate_answers`

src/distilabel/steps/task/evol_instruct/generator.py
```diff
@@ -0,0 +1,244 @@
...
+from distilabel.steps.task.base import GeneratorTask
+from distilabel.steps.task.evol_instruct.utils import GenerationMutationTemplates
+from distilabel.steps.task.typing import ChatType
+
+if TYPE_CHECKING:
+    from distilabel.steps.typing import GeneratorStepOutput
+
+
+class EvolInstructGenerator(GeneratorTask):
+    """WizardLM: Empowering Large Language Models to Follow Complex Instructions
```
gabrielmbmb, line 45: brief description and columns section

src/distilabel/steps/task/evol_instruct/generator.py
gabrielmbmb, line 56: add description to `min_length` and `max_length`

src/distilabel/steps/task/evol_instruct/generator.py
```diff
@@ -0,0 +1,244 @@
...
+                self.mutation_templates.FRESH_START.value.replace(  # type: ignore
+                    "<PROMPT>",
+                    ", ".join(
+                        [
+                            np.random.choice(self._english_nouns).strip()
+                            for _ in range(num_words)
+                        ]
+                    ),
+                )
+            )
```
gabrielmbmb, line 84: i would put this in a method like `_init_seed_texts`

src/distilabel/steps/task/evol_instruct/generator.py
```diff
@@ -0,0 +1,244 @@
...
+                self.output_mappings["answer"]: answer,
+                self.output_mappings["model_name"]: self.llm.model_name,
+            }
+        return {
+            self.output_mappings["instruction"]: instruction,
+            self.output_mappings["model_name"]: self.llm.model_name,
+        }
+
+    @override
+    def process(self) -> "GeneratorStepOutput":  # type: ignore
```
gabrielmbmb, line 138: Same as in the other one, I would split `process` in two methods

gabrielmbmb (changes requested):
Hi @alvarobartt, it looks good to me. Some methods are quite big so I would split them

src/distilabel/steps/task/evol_instruct/base.py
alvarobartt: Yes, note that the initial instruction is already part of the input i.e. `instructions`, all those will be included within `evolved_instructions`, so the starting instruction won't get lost unless explicitly overwritten.

src/distilabel/steps/task/evol_instruct/base.py
```diff
@@ -0,0 +1,213 @@
...
+        """The output for the task is the `generation` and the `model_name`."""
+        # TODO: having to define a `model_name` column every time as the `Task.outputs` is not ideal,
+        # this could be handled always and the value could be included within the DAG validation when
+        # a `Task` is used, since all the `Task` subclasses will have an `llm` with a `model_name` attr.
+        _outputs = [
+            "evolved_instruction"
+            if not self.store_evolutions
+            else "evolved_instructions",
+            "model_name",
+        ]
```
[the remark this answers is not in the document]
alvarobartt: IMO is really confusing to have a column named `outputs` and have a single item, as well as `output` and have a list of items :/ So for the moment I'd like to keep it as is

src/distilabel/steps/task/base.py
```diff
@@ -85,3 +85,7 @@ def process(self, inputs: StepInput) -> "StepOutput":  # type: ignore
             input["model_name"] = self.llm.model_name
             outputs.append(input)
         yield outputs
+
+
+class GeneratorTask(GeneratorStep, Task):
```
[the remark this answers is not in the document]
alvarobartt: Hmm but if the `Task` implements `process` and `load` which are shared for both `Task` and `GeneratorTask` what's the purpose on creating `_Task`? Because what you mentioned above implies `_Task` -> `Task(Step, _Task)` -> `GeneratorTask(GeneratorStep, Task)`, is that correct? Othewise what you mentioned implies that `_Task` contains everything including `load` and `process` and both `Task` and `GeneratorTask` inherit from `_Task` and either `Step` or `GeneratorStep`, and then `pass`.

gabrielmbmb (approved):

PR #415, opened 2024-03-13
Add `EvolComplexity` task

### LINKED ISSUE #392: Adapt `EvolComplexityTask` to new `Task` interface
Adapt the `EvolComplexityTask` to the new `Task` interface added in `1.0.0`. This task will have a runtime argument that will allow to specify a value indicating the number of evolutions to perform.

davidberenstein1957:
## Description

Implements EvolComplexity as a wrapper around EvolInstruct but with limited template options.

Closes #392


alvarobartt (commented):
Didn't review yet, but I'd place `evol_complexity` under `EvolInstruct` since it's based on it, rather than having it in a separate module, we can work on the imports later on i.e. `from distilabel.tasks import EvolComplexity, EvolInstruct`

davidberenstein1957:
@alvarobartt, I thought it was more structured when searching for tasks to have them in separate folders but I'll change the structure a bit.

davidberenstein1957:
@alvarobartt feel free to review

tests/unit/steps/task/evol_instruct/test_generator.py
alvarobartt, line 1: I wouldn't mix the tests here, i.e. `EvolInstruct` separately

src/distilabel/steps/task/evol_instruct/base.py
```diff
@@ -73,7 +76,6 @@ def model_post_init(self, __context: Any) -> None:
         This is useful if you want to do some validation that requires the entire model to be initialized.
         """
         super().model_post_init(__context)
-
         np.random.seed(self.seed)
```
alvarobartt, line 79:
```suggestion

        np.random.seed(self.seed)
```

src/distilabel/steps/task/evol_instruct/base.py
```diff
@@ -261,3 +265,17 @@ def process(self, inputs: "StepInput") -> "StepOutput":  # type: ignore
...
+
+class EvolComplexity(EvolInstruct):
+    """
+    What Makes Good Data for Alignment? A Comprehensive Study of Automatic Data Selection in Instruction Tuning
+    and
+    WizardLM: Empowering Large Language Models to Follow Complex Instructions
+    Reference:
+        - https://arxiv.org/abs/2312.15685
+        - https://arxiv.org/abs/2304.12244
+        - https://github.com/h2oai/h2o-wizardlm
```
alvarobartt, line 278: This reference shouldn't be here right?

```suggestion
```

src/distilabel/steps/task/evol_instruct/base.py
```diff
@@ -261,3 +265,17 @@ def process(self, inputs: "StepInput") -> "StepOutput":  # type: ignore
             for idx, (input, instruction) in enumerate(zip(inputs, instructions)):
                 input.update(self.format_output(instruction, answers[idx]))
             yield inputs
+
+
+class EvolComplexity(EvolInstruct):
```
alvarobartt, line 270: Here we should enhance the docstrings to use the latest formatting

alvarobartt (commented):
Hi here! Some comments, but besides that I meant to nest the `EvolComplexity` within a separate directory as `evol_instruct/complexity/base.py` or something similar, to me is clearer, WDYT?

tests/unit/steps/task/evol_instruct/test_generator.py
davidberenstein1957: I think we'll create a lot of code/test-duplication for inherently testing the same things so I think it is better to keep it like this.

alvarobartt: I was thinking that as `EvolComplexity` is just inheriting from `EvolInstruct` and modifying the `mutation_templates`, only that should be tested. Because the rest of it is just a duplication from `EvolInstruct`, so there's no need to run the same tests twice, so a check on the `mutation_templates` should be enough IMO

alvarobartt (approved):

PR #416, opened 2024-03-13
Add `num_generations` and `group_generations` parameters to `Task` 

gabrielmbmb:
## Description

This PR adds the `num_generations` runtime parameter which will be used to define the number of generations to create per input using an LLM. In addition, the `group_generations` parameter is added which defines whether a list with the `num_generations` is created or if a row with each generation is created:

```python
# num_generations = 3, group_generations = True
[
    {"instruction": "Hello, how are you?", "generation": ["Hi, I'm fine", "Hi, I'm fine, thank you for asking", "Hi! I'm fine and you?"]}
]

# num_generations = 3, group_generations = False
[
    {"instruction": "Hello, how are you?", "generation": "Hi, I'm fine"},
    {"instruction": "Hello, how are you?", "generation": "Hi, I'm fine, thank you for asking"},
    {"instruction": "Hello, how are you?", "generation": "Hi! I'm fine and you?"},
]
```

In addition, this PR has updated all the `LLM`s implementations to handle the `num_generations` parameter.

Closes #413

PR #421, opened 2024-03-14
Add `ComplexityScorer` task

### LINKED ISSUE #391: Adapt `ComplexityScorerTask` to new `Task` interface

gabrielmbmb:
## Description

This PR adds the `ComplexityScorer` task described in the [DEITA paper](https://arxiv.org/abs/2312.15685) (thanks @plaguss for the previous implementation).

🚨 This PR should be merged after merging #416.

Close #391


src/distilabel/steps/task/complexity_scorer.py
```diff
@@ -0,0 +1,86 @@
...
+respond with the format:
+[1] Score: 1
+[2] Score: 2
+...
+{% for instruction in instructions %}
+[{{ loop.index }}] {{ instruction }}
+{%- endfor %}
+""".lstrip()
+
+_PARSE_SCORE_LINE_REGEX = re.compile(r"\[\d+\] score: (\d+)")
```
alvarobartt, line 38: I think ignoring the casing in `score` is safer, also because within the prompt the format is not `[N] score: M` but `[N] Score: M`, and adding `re.IGNORECASE` is more robust in this case.

```suggestion
_PARSE_SCORE_LINE_REGEX = re.compile(r"\[\d+\] score: (\d+)", re.IGNORECASE)
```

src/distilabel/steps/task/complexity_scorer.py
```diff
@@ -0,0 +1,86 @@
...
+    """This task is used to rank a list of instructions based on their complexity. It's
+    an implementation of the complexity score task from the paper 'What Makes Good Data
+    for Alignment? A Comprehensive Study of Automatic Data Selection in Instruction Tuning'.
+
+    Columns:
+    - `inputs`: instructions
+    - `outputs`: complexity_score
+
+    Reference:
+        - https://arxiv.org/abs/2312.15685
```
alvarobartt, line 51: Should we add the reference before the `Columns`? Also the formatting for `Columns` is unclear, I'd proceed as if those were `Args` and use the same formatting instead. But I would also add those at the end, as well as the runtime parameters so that we have `docstring` + our formatting, instead of a mixture.

alvarobartt: i.e. the default docstring format, and then everything else we want to include as extra fields

src/distilabel/steps/task/complexity_scorer.py
gabrielmbmb: In the `process` function we're lowering before applying the regex. We can add this and remove the lower.

ignacioct (approved):
Looks good to me :) Also, thanks a lot for this code, QualityScorer is super similar and it was like a crash course to me.

src/distilabel/steps/task/complexity_scorer.py
```diff
@@ -0,0 +1,88 @@
...
+    for Alignment? A Comprehensive Study of Automatic Data Selection in Instruction Tuning'.
+
+    Input columns:
+        instructions (`List[str]`): The list of instructions to be scored.
+
+    Output columns:
+        complexity_score (`List[float]`): The complexity score for each instruction.
+
+    Reference:
+        - https://arxiv.org/abs/2312.15685
```
ignacioct, line 53:
```suggestion
        - [`What Makes Good Data for Alignment? A Comprehensive Study of Automatic Data Selection in Instruction Tuning`](https://arxiv.org/abs/2312.15685)
```

In the previous task, we used to have this format.

gabrielmbmb: doesn't hurt anyone so I think it's ok! @alvarobartt

PR #432, opened 2024-03-18
Add `api_key` within `LLM.load` and add `llm_kwargs` as `RuntimeParameter`

### LINKED ISSUE #422: Set `api_key` or any other `SecretStr` as `RuntimeParameter`
## Description

The validation for the secret values i.e. `api_key` should be done within the `load` so that the `Pipeline` is not affected when dumped into a file and then loaded back, since the secrets will not be stored for security reasons, so those should be controllable via `RuntimeParameter` not only via `arg` or environment variable.

alvarobartt:
## Description

This PR adds the `llm_kwargs` as `RuntimeParameter` within the `Task` so that those can be propagated to the `LLM.load` method (depending on which ones are exposed).

Additionally, the `api_key` has been included in both `OpenAILLM.load` and `MistralLLM.load` methods, so that it can be defined via `llm_kwargs` from the `Task` so as to solve the existing issue with the `api_key` during the serialization, as it's excluded from it, and then couldn't be defined, as it was only a `__init__` arg.

Finally, also `_handle_api_key_value` method has been included to `LLM` so as to unify the `api_key` handling of the values to the `LLM` base class, as it can be reused on subclasses.

Closes #422

Besides that, also the `_Task` subclassing has been fixed to subclass `Step` in the first place, and then for `GeneratorTask` to subclass from `GeneratorStep` and `_Task`, as otherwise some issues were encountered.

## Example

```python
from distilabel.steps.task.text_generation import TextGeneration

if __name__ == "__main__":
    with Pipeline() as pipeline:
        ...

        openai_generation = TextGeneration(
            name="openai_generation",
            llm=OpenAILLM(),
        )
        
        ...

    pipeline.run(
        parameters={
            ...,
            "openai_generation": {
                "llm_kwargs": {
                    "api_key": "sk-***",
                },
                "generation_kwargs": {
                    "max_new_tokens": 512,
                    "temperature": 0.7,
                },
            },
        }
    )
```

### DIFF
--- a/src/distilabel/llm/base.py
+++ b/src/distilabel/llm/base.py
@@ -17,7 +17,7 @@ import logging
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, Dict, List, Union
 
-from pydantic import BaseModel, ConfigDict, PrivateAttr
+from pydantic import BaseModel, ConfigDict, PrivateAttr, SecretStr
 
 from distilabel.utils.logging import get_logger
 from distilabel.utils.serialization import _Serializable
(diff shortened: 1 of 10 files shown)


dvsrepo:
> ## Description
> This PR adds the `llm_kwargs` as `RuntimeParameter` within the `Task` so that those can be propagated to the `LLM.load` method (depending on which ones are exposed).
> 
> Additionally, the `api_key` has been included in both `OpenAILLM.load` and `MistralLLM.load` methods, so that it can be defined via `llm_kwargs` from the `Task` so as to solve the existing issue with the `api_key` during the serialization, as it's excluded from it, and then couldn't be defined, as it was only a `__init__` arg.
> 
> Finally, also `_handle_api_key_value` method has been included to `LLM` so as to unify the `api_key` handling of the values to the `LLM` base class, as it can be reused on subclasses.
> 
> Closes #422
> 
> Besides that, also the `_Task` subclassing has been fixed to subclass `Step` in the first place, and then for `GeneratorTask` to subclass from `GeneratorStep` and `_Task`, as otherwise some issues were encountered.
> 
> ## Example
> ```python
> from distilabel.llm.openai import OpenAILLM
> from distilabel.pipeline.local import Pipeline
> from distilabel.steps.task.text_generation import TextGeneration
> 
> if __name__ == "__main__":
>     with Pipeline() as pipeline:
>         ...
> 
>         openai_generation = TextGeneration(
>             name="openai_generation",
>             llm=OpenAILLM(),
>         )
>         
>         ...
> 
>     pipeline.run(
>         parameters={
>             ...,
>             "openai_generation": {
>                 "llm_kwargs": {
>                     "api_key": "sk-***",
>                 },
>                 "generation_kwargs": {
>                     "max_new_tokens": 512,
>                     "temperature": 0.7,
>                 },
>             },
>         }
>     )
> ```

Hi! I don't have the full picture but from a user standpoint this way of passing/configuring the API key feels very unnatural and overly complicated. I'm sure there's a reason but we should analyze the trade-offs because I would expect to be able to setup this when I instantiate the LLM, not with a complex dict structure when I run the pipeline. I'm just looking at the code snippet so I'm missing if this just an alternative way of doing things and not the default/only way, in which case it's fine and you can forget about my comment

alvarobartt:
> Hi! I don't have the full picture but from a user standpoint this way of passing/configuring the API key feels very unnatural and overly complicated. I'm sure there's a reason but we should analyze the trade-offs because I would expect to be able to setup this when I instantiate the LLM, not with a complex dict structure when I run the pipeline. I'm just looking at the code snippet so I'm missing if this just an alternative way of doing things and not the default/only way, in which case it's fine and you can forget about my comment

Hi here @dvsrepo! Indeed this is just an alternative way, one can do e.g. `OpenAILLM(api_key="...")` also setting `OPENAI_API_KEY` and then `OpenAILLM()` or just providing those in runtime! And the scenario for that is when restoring a `Pipeline` from a previously serialized one, and wanting to provide those within the `pipeline.run` method rather than via environment variable, as during the `Pipeline` deserialization all the components are automatically instantiated so the `__init__` is called internally with the existing args and cannot be overriden on `Pipeline` loading time!

dvsrepo:
Yipiii, thanks @alvarobartt

PR #435, opened 2024-03-18
Add `GeneratorStep.process` validation in `DAG` and smaller fixes

### LINKED ISSUE #431: Raise `RuntimeError` if `GeneratorStep.process` signature contains `inputs`
## Description

For a more meaningful logging, we should try to catch all the possible issues, and one should be to check the method signatures before trying to serialize internally the steps. As e.g. `GeneratorStep.process` won't accept `inputs` as arg, since it will generate data out of no inputs, so adding `inputs` will end up in a `thread.RLock` issue that's hard to debug without prior context if users instantiate the `GeneratorStep` subclass poorly.

P.S. Additionally, all these things should be crystal clear in the documentation too!

alvarobartt:
## Description

This PR adds a method within `DAG` named `_validate_generator_step_process_signature` that validates that the provided `process` method signature matches the expected one, which should contain the `offset` arg, and also the `inputs` arg should not be there, otherwise, a `ValueError` will be raised during the validation.

Closes #431

Additionally, this PR also tackles the following:

* Add missing `CombineColumns` docstrings
* Use `Yields` over `Returns` in docstrings where `yield` is used instead of `return`
* Move `Task` specific methods out of abstract `_Task`, to avoid conflicts / collisions with `GeneratorTask` that has a different `process` signature

### DIFF
--- a/src/distilabel/integrations/argilla/base.py
+++ b/src/distilabel/integrations/argilla/base.py
@@ -26,12 +26,11 @@ except ImportError as ie:
         "Argilla is not installed. Please install it using `pip install argilla`."
     ) from ie
 
-from distilabel.steps.base import Step
+from distilabel.steps.base import Step, StepInput
 
 if TYPE_CHECKING:
     from argilla.client.feedback.dataset.remote.dataset import RemoteFeedbackDataset
 
-    from distilabel.steps.base import StepInput
     from distilabel.steps.typing import StepOutput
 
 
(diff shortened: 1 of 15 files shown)


gabrielmbmb:
Hi, LGTM! I made some small changes to reuse `Step` logic instead of calling `inspect.signature` again, and to check there is no parameter type hinted with `StepInput`, as that's the important thing not the name. BTW, it's important to not import `StepInput` within a `TYPE_CHECKING` block otherwise we cannot check it's a parameter annotated with the metadata we need.

gabrielmbmb (approved):

PR #444, opened 2024-03-19
Add `AnthropicLLM`

### LINKED ISSUE #424: Add `AnthropicLLM`

sdiazlor:
Closes #424 

```python
from distilabel.llm.anthropic import AnthropicLLM

if __name__ == "__main__":
    llm = AnthropicLLM(model="claude-3-opus-20240229", api_key="api.key")
    llm.load()
    output = llm.generate(
        inputs=[
            [
                {
                    "role": "user",
                },
            ]
        ]
    )
    print(output)
```


src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,118 @@
...
+        api_key: the API key to authenticate the requests to the Anthropic API.
+        base_url: the base URL to use for the Anthropic API. Defaults to "https://api.anthropic.com".
+        http_client: the HTTP client to use for the Anthropic API. Defaults to None.
+        timeout: the maximum time in seconds to wait for a response. Defaults to 600.0.
+        max_retries: the maximum number of retries for the LLM. Defaults to 2.
+    """
+
+    model: str = "claude-3-opus-20240229"
+    api_key: Optional[SecretStr] = os.getenv("ANTHROPIC_API_KEY", None)  # type: ignore
+    base_url: str = "https://api.anthropic.com"
```
davidberenstein1957, line 43: Can't we set this to `None` and let this be handled internally?

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,118 @@
...
+    timeout: float = 600.0
+    http_client: Union[str, None] = None
```
davidberenstein1957, line 45: Probably the `http_client` isn't a `str`-type?

pyproject.toml
```diff
@@ -42,6 +42,7 @@ openai = ["openai >= 1.0.0"]
 vllm = ["vllm >= 0.2.1"]
 vertexai = ["google-cloud-aiplatform >= 1.38.0"]
 together = ["together"]
+anthropic =["anthropic >= 0.20.0"]
```
alvarobartt, line 45: This should be ordered alphabetically i.e. `anthropic` and `argilla` should go on top of the list of extras (I believe the rest are more or less sorted, but feel free to do so for the rest too)

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,118 @@
...
+    from distilabel.steps.task.typing import ChatType
+
+
+class AnthropicLLM(AsyncLLM):
+    """Anthropic LLM implementation running the Async API client.
+
+    Args:
+        model: the model name to use for the LLM.
+        api_key: the API key to authenticate the requests to the Anthropic API.
+        base_url: the base URL to use for the Anthropic API. Defaults to "https://api.anthropic.com".
```
alvarobartt, line 35:
```suggestion
        base_url: the base URL to use for the Anthropic API. Defaults to `None` which means that `https://api.anthropic.com` will be used internally.
```

src/distilabel/llm/anthropic.py
alvarobartt: Note that all the types within the `pydantic.BaseModel` need to be serializable i.e. `httpx.Client` here would fail unless a `PrivateAttr`

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,118 @@
...
+        try:
+            from anthropic import AsyncAnthropic
+        except ImportError as ie:
+            raise ImportError(
+                "Anthropic Python client is not installed. Please install it using"
+                " `pip install anthropic`."
+            ) from ie
+
+        self.api_key = self._handle_api_key_value(
+            self_value=self.api_key, load_value=api_key, env_var="ANTHROPIC_API_KEY"
```
alvarobartt, line 61: Could you set a `PrivateAttr` named `_env_var` with the `ANTHROPIC_API_KEY` as in https://github.com/argilla-io/distilabel/blob/1e52450400a0b324798268048ffa06da0297181a/src/distilabel/llm/openai.py#L43

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,118 @@
...
+        stop_sequences: List[str] = None,
+        temperature: float = 1.0,
+        top_p: float = 1.0,
+        top_k: int = 0,
+    ) -> "GenerateOutput":
+        """Generates a response asynchronously, using the [Anthropic Async API definition](https://github.com/anthropics/anthropic-sdk-python).
+
+        Args:
+            input: a single input in chat format to generate responses for.
+            system: the system prompt to use for the generation. No existing 'system' role. Defaults to `""`.
```
alvarobartt, line 92:
```suggestion
            system: the system prompt to use for the generation. No existing `system` role within the input conversation, only `user` and `assistant`. Defaults to `""`.
```

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,118 @@
...
+            temperature: the temperature to use for the generation. Set only if top_p is None. Defaults to `1.0`.
+            top_p: the top-p value to use for the generation. Defaults to `1.0`.
+            top_k: the top-k value to use for the generation. Defaults to `0`.
+
+        Returns:
+            A list of lists of strings containing the generated responses for each input.
+        """
+
+        generations = []
+        for _ in range(num_generations):
```
alvarobartt, line 105: I believe we discussed internally that for the `LLM`s not supporting `n` within the API call we were not going to implement this for-loop, can you confirm @gabrielmbmb?

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,118 @@
...
+                system=system,
+                messages=input,
+                max_tokens=max_tokens,
+                stream=False,
+                stop_sequences=stop_sequences,
+                temperature=temperature,
+                top_p=top_p,
+                top_k=top_k,
+            )
+            generations.append(completion.content[0].text)
```
alvarobartt, line 117: For handling the generations feel free to take some inspiration from https://github.com/argilla-io/distilabel/blob/1e52450400a0b324798268048ffa06da0297181a/src/distilabel/llm/openai.py#L115-L123

tests/unit/llm/test_anthropic.py
```diff
@@ -0,0 +1,89 @@
...
+from unittest.mock import AsyncMock, MagicMock, Mock, patch
+
+import nest_asyncio
+import pytest
+from distilabel.llm.anthropic import AnthropicLLM
+
+
+@patch("anthropic.AsyncAnthropic")
+class TestAnthropicLLM:
+    def test_anthropic_llm(self, mock_anthropic: MagicMock) -> None:
```
alvarobartt, line 25:
```suggestion
    def test_anthropic_llm(self) -> None:
```

alvarobartt (commented):
Hi @sdiazlor! Thanks for the PR, there's a couple of things I've mentioned and some others I'd like to confirm as:

* Is there any way we can check whether the provided model is an existing one? Could we offer a listing or just call an API endpoint to get the list of existing models within Anthropic?
* Could you reply with an example using the `http_client`? Or just mentioning which are the scenarios where it's useful? Also note that there may be issues when serializing the `http_client` if that appears to expect an `httpx.Client` instance

tests/unit/llm/test_anthropic.py
alvarobartt: Hi here @sdiazlor this feedback was wrong, I think you need to explicitly provide it, my bad!

tests/unit/llm/test_anthropic.py
```diff
@@ -0,0 +1,89 @@
...
+            system="",
+            inputs=[
+                {
+                    "role": "user",
+                    "content": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+                },
+            ],
+        )
+
+    def test_serialization(self, mock_anthropic: MagicMock) -> None:
```
davidberenstein1957, line 72: if `mock_anthropic` is not user, you can replace it with `_`

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @property
+    def model_name(self) -> str:
+        """Returns the model name used for the LLM."""
+        return self.model
+
+    async def agenerate(  # type: ignore
+        self,
+        input: "ChatType",
+        system: str = "",
```
gabrielmbmb, line 83: We can remove this parameter, as it would be part of `input`:

```python
[
    {"role": "system", "content": "..."},
    {"role": "user", "content": "..."}
]
```

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,124 @@
...
+            top_p: the top-p value to use for the generation. Defaults to `1.0`.
+            top_k: the top-k value to use for the generation. Defaults to `0`.
+
+        Returns:
+            A list of lists of strings containing the generated responses for each input.
+        """
+
+        completion = await self._aclient.messages.create(
+            model=self.model,
+            system=system,
```
gabrielmbmb, line 109: try to get the system prompt from the first element of `input` if role == system. Otherwise, leave client default value.

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,124 @@
...
+            stop_sequences=stop_sequences,
+            temperature=temperature,
+            top_p=top_p,
+            top_k=top_k,
+        )
+        if (content := completion.content[0].text) is None:
+            self._logger.warning(
+                f"Received no response using Anthropic client (model: '{self.model}')."
+                f" Finish reason was: {completion.stop_reason}"
+            )
```
gabrielmbmb, line 122: As Anthropic AI client doesn't support an `n` or `num_generations` parameter, we should handle that on distilabel implementation overriding `generate` method like here: 

https://github.com/argilla-io/distilabel/blob/358333c6aa3b5ce8d9f55cf0e1f58c87cb59cdd7/src/distilabel/llm/mistral.py#L140-L164

gabrielmbmb (changes requested):
Hi, it looks good!. Just need a small update before merging.

src/distilabel/llm/anthropic.py
```diff
@@ -0,0 +1,124 @@
...
+        """Returns the model name used for the LLM."""
+        return self.model
+
+    async def agenerate(  # type: ignore
+        self,
+        input: "ChatType",
+        system: str = "",
+        num_generations: int = 1,
+        max_tokens: int = 128,
+        stop_sequences: List[str] = None,
```
gabrielmbmb, line 86:
```suggestion
        stop_sequences: Union[List[str], None] = None,
```

gabrielmbmb (approved):

PR #445, opened 2024-03-19
Add `VertexAILLM`  

### LINKED ISSUE #378: Adapt `VertexAILLM` to new `LLM` interface

davidberenstein1957:
closes #378 

- I only implemented gemini models, which allowed for the openai-like chat format and are mostly likely used the most.
- the SDK and API don't align with `openai` standards and expect a `user` or `model` role. They don't support `assistant` or `system`

```python 
import time

import vertexai
from distilabel.llm.vertexai import VertexAILLM

project_id = FILL_OUT_YOURSELF
location = "europe-west1"
vertexai.init(project=project_id)

if __name__ == "__main__":

    start_time = time.time()
    llm = VertexAILLM()
    llm.load()
    output = llm.generate(
        inputs=[
            [
                {
                    "role": "user",
                },
            ]
        ]
    )
    exit()
```


src/distilabel/llm/vertexai.py
```diff
@@ -0,0 +1,173 @@
...
+            self._content_class = Content
+            self._part_class = Part
+            self._generation_config_class = GenerationConfig
+        except ImportError as e:
+            raise ImportError(
+                "vertexai is not installed. Please install it using"
+                " `pip install google-cloud-aiplatform`."
+            ) from e
+        except Exception as e:
+            raise e
```
gabrielmbmb, line 91: is this one needed?

src/distilabel/llm/vertexai.py
```diff
@@ -0,0 +1,173 @@
...
+
+        if _is_gemini_model(self.model):
+            self._aclient = GenerativeModel(model_name=self.model)
+        else:
+            raise NotImplementedError(
+                "`VertexAILLM` is only implemented for `gemini` models that allow for `ChatType`data."
```
gabrielmbmb, line 97:
```suggestion
                "`VertexAILLM` is only implemented for `gemini` models that allow for `ChatType` data."
```

src/distilabel/llm/vertexai.py
```diff
@@ -0,0 +1,173 @@
...
+            tools: A potential list of tools that can be used by the API. Defaults to `None`.
+
+        Returns:
+            A list of lists of strings containing the generated responses for each input.
+        """
+        contents = self._chattype_to_content(input)
+        content = await self._aclient.generate_content_async(
+            contents=contents,
+            generation_config=self._generation_config_class(
+                candidate_count=num_generations,
```
gabrielmbmb, line 161: Do you know if they have updated the `candidate_count` parameter and now more than 1 generation is allowed?

src/distilabel/llm/vertexai.py
```diff
@@ -0,0 +1,173 @@
...
+                top_p=top_p,
+                max_output_tokens=max_output_tokens,
+                stop_sequences=stop_sequences,
+            ),
+            safety_settings=safety_settings,
+            tools=tools,
+            stream=False,
+        )
+
+        return [candidate.text for candidate in content.candidates]
```
gabrielmbmb, line 173: Here when trying to access `text`, it can raise a `ValueError`. We should catch that error and impute the generation with `None`. Also log a warning message with the finish reason like here: https://github.com/argilla-io/distilabel/blob/358333c6aa3b5ce8d9f55cf0e1f58c87cb59cdd7/src/distilabel/llm/openai.py#L140-L147

You can get the finish reason accessing `finish_reason` attribute of the `Candidate` instance.

gabrielmbmb (changes requested):
LGTM, just a small details before merging

src/distilabel/llm/vertexai.py
davidberenstein1957: Nope, I though I had removed this already.

src/distilabel/llm/vertexai.py
davidberenstein1957: nope, I assumed it was.

PR #439, opened 2024-03-19
Add `InferenceEndpointsLLM`

alvarobartt:
## Description

This PR adds the `InferenceEndpointsLLM` back, powered by the OpenAI client, and running on both serverless and dedicated inference endpoints running TGI.

Closes #375

## Example

```python
...

if __name__ == "__main__":
    start_time = time.time()

    with Pipeline() as pipeline:
        load_hub_dataset = LoadHubDataset(name="load_dataset", batch_size=5)

        evol_instruct = EvolInstruct(
            name="evol_instruct",
            llm=InferenceEndpointsLLM(
                base_url="<INFERENCE_ENDPOINTS_URL>",
                api_key="hf_***",
            ),
            num_evolutions=5,
            store_evolutions=True,
            generate_answers=True,
        )
        load_hub_dataset.connect(evol_instruct)

        push_to_hub = PushToHub(name="push_to_hub")
        evol_instruct.connect(push_to_hub)

    pipeline.run(
        parameters={
            "load_dataset": {
                "repo_id": "<INPUT_REPO_ID>",
                "split": "train",
            },
            "evol_instruct": {
                "generation_kwargs": {
                    "max_new_tokens": 512,
                    "temperature": 0.7,
                },
            },
            "push_to_hub": {
                "repo_id": "<OUTPUT_REPO_ID>",
                "split": "train",
                "private": False,
                "token": "hf_***",
            },
        }
    )

    print("--- %s seconds ---" % (time.time() - start_time))
```

PR #447, opened 2024-03-19
446 anyscalellm

### LINKED ISSUE #446: Add `AnyScaleLLM` integration

davidberenstein1957:
Closes #446 

```python
import time

from distilabel.llm.anyscale import AnyscaleLLM

if __name__ == "__main__":

    start_time = time.time()
    llm = AnyscaleLLM(api_key="esecret_*****")
    llm.load()
    output = llm.generate(
        inputs=[
            [
                {
                    "role": "user",
                },
            ]
        ]
    )
    exit()
```

### DIFF
--- a/src/distilabel/llm/anyscale.py
+++ b/src/distilabel/llm/anyscale.py
@@ -25,13 +25,14 @@ class AnyscaleLLM(OpenAILLM):
         model: the model name to use for the LLM e.g. "gpt-3.5-turbo", "gpt-4", etc.
         base_url: the base URL to use for the Anyscale API can be set with `OPENAI_BASE_URL`. Default is "https://api.endpoints.anyscale.com/v1".
         api_key: the API key to authenticate the requests to the Anyscale API. Can be set with `OPENAI_API_KEY`. Default is `None`.
-
     """
 
     model: str = "mistralai/Mixtral-8x7B-Instruct-v0.1"
     base_url: Optional[str] = "https://api.endpoints.anyscale.com/v1"
 
-    def load(self, api_key: Optional[str] = None) -> None:
+    def load(
+        self, api_key: Optional[str] = None, base_url: Optional[str] = None
+    ) -> None:
         """Loads the `AsyncOpenAI` client to benefit from async requests."""
 
         try:
@@ -46,10 +47,11 @@ class AnyscaleLLM(OpenAILLM):
             self_value=self.api_key, load_value=api_key, env_var="OPENAI_API_KEY"
         )
         self.base_url = self._handle_api_key_value(
-            self_value=self.api_key, load_value=api_key, env_var="OPENAI_BASE_URL"
+            self_value=self.base_url, load_value=base_url, env_var="OPENAI_BASE_URL"
         )
 
         self._aclient = AsyncOpenAI(
             api_key=self.api_key.get_secret_value(),
+            base_url=self.base_url.get_secret_value(),
             max_retries=6,
         )
--- a/src/distilabel/llm/anyscale.py
+++ b/src/distilabel/llm/anyscale.py
@@ -1,0 +1,55 @@
+# Copyright 2023-present, Argilla, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+from distilabel.llm.openai import OpenAILLM
+
+
+class AnyscaleLLM(OpenAILLM):
+    """
+    Anyscale LLM implementation running the async API client of OpenAI because of duplicate API behavior.
+
+    Attributes:
+        model: the model name to use for the LLM e.g. "gpt-3.5-turbo", "gpt-4", etc.
+        base_url: the base URL to use for the Anyscale API can be set with `OPENAI_BASE_URL`. Default is "https://api.endpoints.anyscale.com/v1".
+        api_key: the API key to authenticate the requests to the Anyscale API. Can be set with `OPENAI_API_KEY`. Default is `None`.
+
+    """
+
+    model: str = "mistralai/Mixtral-8x7B-Instruct-v0.1"
+    base_url: Optional[str] = "https://api.endpoints.anyscale.com/v1"
+
+    def load(self, api_key: Optional[str] = None) -> None:
+        """Loads the `AsyncOpenAI` client to benefit from async requests."""
+
+        try:
+            from openai import AsyncOpenAI
+        except ImportError as ie:
+            raise ImportError(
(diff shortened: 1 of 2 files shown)


src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Anyscale LLM implementation running the async API client of OpenAI because of duplicate API behavior.
+
+    Attributes:
+        model: the model name to use for the LLM e.g. "gpt-3.5-turbo", "gpt-4", etc.
+        base_url: the base URL to use for the Anyscale API can be set with `OPENAI_BASE_URL`. Default is "https://api.endpoints.anyscale.com/v1".
+        api_key: the API key to authenticate the requests to the Anyscale API. Can be set with `OPENAI_API_KEY`. Default is `None`.
+    """
+
+    model: str = "mistralai/Mixtral-8x7B-Instruct-v0.1"
```
alvarobartt, line 30: I wouldn't set any default `model`, plus a check on whether the provided model exists or not within Anyscale would be nice! i.e. calling an API endpoint to list the available text generation models

```suggestion
    model: str
```

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,57 @@
...
+
+from distilabel.llm.openai import OpenAILLM
+
+
+class AnyscaleLLM(OpenAILLM):
+    """
+    Anyscale LLM implementation running the async API client of OpenAI because of duplicate API behavior.
+
+    Attributes:
+        model: the model name to use for the LLM e.g. "gpt-3.5-turbo", "gpt-4", etc.
```
alvarobartt, line 25: I believe this docstring needs an update

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,57 @@
...
+        base_url: the base URL to use for the Anyscale API can be set with `OPENAI_BASE_URL`. Default is "https://api.endpoints.anyscale.com/v1".
+        api_key: the API key to authenticate the requests to the Anyscale API. Can be set with `OPENAI_API_KEY`. Default is `None`.
+    """
+
+    model: str = "mistralai/Mixtral-8x7B-Instruct-v0.1"
+    base_url: Optional[str] = "https://api.endpoints.anyscale.com/v1"
```
alvarobartt, line 31:
```suggestion
    base_url: str = "https://api.endpoints.anyscale.com/v1"
```

alvarobartt, line 31: In #439 I've included the `base_url` as an attribute in `OpenAILLM` so feel free to reuse that once we merge #439

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,57 @@
...
+                "OpenAI Python client is not installed which is required for `AnyscaleLLM`. Please install it using"
+                " `pip install openai`."
+            ) from ie
+
+        self.api_key = self._handle_api_key_value(
+            self_value=self.api_key, load_value=api_key, env_var="OPENAI_API_KEY"
+        )
+        self.base_url = self._handle_api_key_value(
+            self_value=self.base_url, load_value=base_url, env_var="OPENAI_BASE_URL"
+        )
```
alvarobartt, line 51: Even though `OPENAI_...` environment variables can be reused for this, maybe we're safe just assuming that the API key will be provided via `ANYSCALE_API_KEY`? Also what are the benefits of providing the `base_url` via environment variable if it's always going to be the default Anyscale URL?

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,57 @@
...
+
+        self._aclient = AsyncOpenAI(
+            api_key=self.api_key.get_secret_value(),
+            base_url=self.base_url.get_secret_value(),
```
alvarobartt, line 55: Why would the `base_url` had to be a `SecretStr`? I believe this is due to re-using the function `_handle_api_key_value` but probably we can just assume this is an attribute with no environment variable counterpart, and public.

tests/unit/llm/test_anyscale.py
```diff
@@ -0,0 +1,44 @@
...
+
+import os
+from unittest.mock import MagicMock, patch
+
+from distilabel.llm.anyscale import AnyscaleLLM
+
+
+@patch("openai.AsyncOpenAI")
+class TestAnyscaleLLM:
+    def test_openai_llm(self, mock_openai: MagicMock) -> None:
```
alvarobartt, line 23:
```suggestion
    def test_anyscale_llm(self) -> None:
```

tests/unit/llm/test_anyscale.py
```diff
@@ -0,0 +1,44 @@
...
+        llm = AnyscaleLLM(
+            model="mistralai/Mixtral-8x7B-Instruct-v0.1", api_key="api.key"
+        )  # type: ignore
+        assert isinstance(llm, AnyscaleLLM)
+        assert llm.model_name == "mistralai/Mixtral-8x7B-Instruct-v0.1"
+
+    def test_serialization(self, mock_openai: MagicMock) -> None:
```
alvarobartt, line 30:
```suggestion
    def test_serialization(self) -> None:
```

alvarobartt (commented):
Thanks for the PR @davidberenstein1957!

src/distilabel/llm/anyscale.py
davidberenstein1957: Within the anyscale docs they reference `ANYSCALE_API_KEY` and recommend setting it via the environment variable so I copied that behavior. Maybe they've got this for on_prem deployments. Given that it is optional I want to keep it like this.

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,54 @@
...
+
+from distilabel.llm.openai import OpenAILLM
+
+
+class AnyscaleLLM(OpenAILLM):
+    """
+    Anyscale LLM implementation running the async API client of OpenAI because of duplicate API behavior.
+
+    Attributes:
+        model: the model name to use for the LLM, e.g., `google/gemma-7b-it`. [Supported models](https://docs.endpoints.anyscale.com/text-generation/supported-models/google-gemma-7b-it).
```
alvarobartt, line 25:
```suggestion
        model: the model name to use for the LLM, e.g., `google/gemma-7b-it`. [Supported models](https://docs.endpoints.anyscale.com/text-generation/supported-models).
```

src/distilabel/llm/anyscale.py
```diff
@@ -0,0 +1,54 @@
...
+        self.api_key = self._handle_api_key_value(
+            self_value=self.api_key, load_value=api_key, env_var="OPENAI_API_KEY"
+        )
+        self.base_url = base_url or self.base_url
+
+        self._aclient = AsyncOpenAI(
+            api_key=self.api_key.get_secret_value(),
+            base_url=self.base_url,
+            max_retries=6,
+        )
```
alvarobartt, line 54: No need to define this function, as there's already an update on `core-refactor`, so feel free to merge the changes from that branch and remove this function instead 👍🏻

alvarobartt (commented):
LGTM, great work! Thanks for also removing the `model` defaults, feel free to merge `core-refactor` in your branch, delete the `load` method from `AnyscaleLLM` and then we can merge, thanks again!

src/distilabel/llm/anyscale.py
davidberenstein1957: Sadly this is not defined haha

PR #449, opened 2024-03-19
Add `TogetherLLM`

davidberenstein1957:
Add `test_together`

Closses #448


alvarobartt (commented):
Thanks for this PR @davidberenstein1957! Similar feedback to the one provided at #447, plus would love to see a more complex pipeline on how well these LLMs behave on more complex pipelines (in case you didn't do that already)

davidberenstein1957:
@alvarobartt I tested a more complex pipeline but did copy it here. I think that we in general need integration tests for `tasks` too.

PR #461, opened 2024-03-21
Improve azure openai support

BramVanroy:
- Adds instructions how to use OpenAI services through Azure in README
- Adds a minor typing improvement for an Azure client
- Skip model checking for Azure models because the Azure client is a weird thing: instead of a model like "gpt4-0123" it requires the deployment name. So `self.model` can be any arbitrary string, so verifying it against supported models won't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ In addition, the following extras are available:
 
 - `hf-transformers`: for using models available in [transformers](https://github.com/huggingface/transformers) package via the `TransformersLLM` integration.
 - `hf-inference-endpoints`: for using the [HuggingFace Inference Endpoints](https://huggingface.co/inference-endpoints) via the `InferenceEndpointsLLM` integration.
-- `openai`: for using OpenAI API models via the `OpenAILLM` integration.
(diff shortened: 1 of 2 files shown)


alvarobartt:
Hi @BramVanroy thanks for this! We're about to close the refactor in the `core-refactor` branch, everything's new and refactored there, so we can either merge yours in `main` and adapt that ourselves to the refactored version, or you can do that yourself, whatever you feel is better 👍🏻  It should be straight forward to extend the `OpenAILLM` class to create a subclass for `AzureOpenAILLM`, but ping us if you have any doubt and willing to work on top of `core-refactor`, otherwise, we'll be happy to review this one! Thanks again

BramVanroy:
Hi @alvarobartt I do not have the time to work on this so preferably the path that takes the least effort from my side. :)

If you plan to write a subclass for OpenAI specifically for Azure, that could be even better because then users can pass in their api_version, endpoint, deployment name, etc. rather than have to pre-define the azure client.

alvarobartt (approved):

PR #463, opened 2024-03-22
Add `use_cache` to `BasePipeline`

### LINKED ISSUE #460: Add `use_cache` constructor parameter to `BasePipeline`
Add `use_cache` parameter to `BasePipeline.__init__` to allow users specify if they want to load the `_BatchManager` from the cache.

plaguss:
## Description

This PR adds an argument `use_cache` to `BasePipeline` to decide whether we should cache the pipeline and reuse it or not.

Closes #460.

### DIFF
--- a/src/distilabel/pipeline/base.py
+++ b/src/distilabel/pipeline/base.py
@@ -90,9 +90,12 @@ class BasePipeline(_Serializable):
         dag: The `DAG` instance that represents the pipeline.
     """
 
-    def __init__(self, cache_dir: Optional["PathLike"] = None) -> None:
+    def __init__(
+        self, cache_dir: Optional["PathLike"] = None, use_cache: bool = True
+    ) -> None:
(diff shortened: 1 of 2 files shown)


src/distilabel/pipeline/base.py
```diff
@@ -255,6 +264,9 @@ def _cache_location(self) -> CacheLocation:
 
     def _cache(self) -> None:
         """Saves the `BasePipeline` using the `_cache_filename`."""
+        if not self._use_cache:
+            return
```
gabrielmbmb, line 268: I think we can still cache the pipeline execution even if `use_cache` is false. The parameter is more about using the cache or not when running the pipeline.

gabrielmbmb (changes requested):
LGTM! just a small comment

gabrielmbmb (approved):

PR #476, opened 2024-03-25
Update `docs` for `distilabel` v1.0 with `mkdocs-material`

plaguss:
## Description

WIP.

This PR updates the documentation for the new codebase.


docs/sections/learn/steps/index.md
```diff
@@ -0,0 +1,134 @@
+# Steps
+
+The [`Step`][distilabel.steps.base.Step] is the base class in `distilabel`, every unit of work in a `Pipeline` will inherit from it.
+
+## What's a Step in distilabel?
+
+This a base class is in charge of processing data, which in the end will be lists of dictionaries. In order to process, it will define the data, it defines two properties: `inputs` and `outputs`, which are a list of strings that represent the names of the columns that the step needs as input or output respectively.
```
gabrielmbmb, line 7:
```suggestion
It's a base class is in charge of processing data, which in the end will be lists of dictionaries. In order to process, it defines two properties: `inputs` and `outputs`, which are a list of strings that represent the names of the columns that the step needs as input or output respectively i.e. the keys that have to be present in the list of dictionaries received by the step and the keys that will be added to these dictionaries after running it.
```

docs/sections/learn/steps/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+```python
+conversation_template
+# ConversationTemplate(name='my-conversation-template', input_mappings={}, output_mappings={}, input_batch_size=50)
+```
+
+The `name` of the `Step`, a mandatory field to identify the `Step` within the `Pipeline`. `input_mappings`, which is a dictionary that can be useful when the names of the output columns of the current step differ from the names of the input columns of the following step/s, and `output_mappings` which can be used to verify the output columns of this step exist after processing the data. If this two fields are left to the default (without content), the different steps will assume the column names along the different steps will be present, and raise an error if that's not the case. And `input_batch_size` (by default set to 50), which is independent for every step and will determine how often it can be executed. If won't matter that much in this step, but as we will see later, other types of steps will come with an `LLM`, so having this flexibility will be really useful.
```
gabrielmbmb, line 62:
```suggestion
- The `name` of the `Step`, a mandatory field to identify the `Step` within the `Pipeline`. 
- `input_mappings`, which is a dictionary that can be useful to map keys from the input dictionaries to the keys expected by the step. For example, if `input_mappings={"instruction": "prompt"}`, that means that the key `prompt` from the input dictionaries will be used as the key `instruction` for the step.
- `output_mappings`, which is a dictionary that can be used to map the `outputs` of the step to other names. For example, if `output_mappings={"conversation": "prompt"}`, that means that the key `conversation` generated by the step will be renamed to `prompt` and the output dictionaries of this step will contain a key called `prompt` instead of `conversation`.
- `input_batch_size` (by default set to 50), which is independent for every step and will determine how many input dictionaries will process at once. If won't matter that much in this step, but as we will see later, other types of steps will come with an `LLM`, so having this flexibility will be really useful.
```

docs/sections/learn/steps/index.md
```diff
@@ -0,0 +1,134 @@
...
+This is a small type step that shows what to expect when we are creating our `Step` objects, which can start from something as simple as generating a conversation template from some columns on a dataset.
+
+## step decorator
+
+If all that we want to apply in a step is some simple processing, it can be easier to just create a plain function, and decorate it. We can find more examples in the [API reference][distilabel.steps.decorator], but let's see how we could define the previous step as a function and use the decorator:
+
+```python
+from distilabel.steps.decorator import step
+from distilabel.steps.typing import StepOutput
+from distilabel.steps.base import StepInput
```
gabrielmbmb, line 99:
```suggestion
from distilabel.steps import StepInput, StepOutput, step 
```

docs/sections/learn/steps/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+Other than the general or normal steps we have seen, there are special types of steps that have a restricted behaviour compared to the general `Step`.
+
+### Generator steps
+
+These are steps that are able to generate data, and don't need to receive any input from previous step, as it's implied in the normal steps. The typical use for these steps will be loading data for example, as can be seen in [`LoadDataFromDicts`][distilabel.steps.generators.data]. For this type of steps we will only need to define the `process` method, and we can optionally pass a `batch_size` argument, that will determine the batch size of the generated batches.
+
+### Global steps
+
+Other special type of step are the global steps. These steps don't have any `inputs` or `outputs`, and their `process` method receives all the data at once instead of using batches. This kind of behavior is necessary for example to push a dataset to a specific place, or doing some filtering on the whole data before continuing with the pipeline.
```
gabrielmbmb, line 133:
```suggestion
Other special type of step are the global steps. The only difference with respect to the `Step` is that their `process` method receives all the data at once instead of using batches. This kind of behaviour is necessary for example to push a dataset to a specific place, or doing some filtering on the whole data before continuing with the pipeline.
```

docs/sections/learn/steps/index.md
gabrielmbmb, line 1: We need to also mention here in the `RuntimeParameter`s

gabrielmbmb (approved):

PR #482, opened 2024-03-26
Move Enum to Dict[str, str] to avoid serialization errors during caching

plaguss:
## Description

This PR removes the following error y changing the `Enum` to `Dict[str, str]` in the `EvolInstruct` family of tasks:

image


src/distilabel/steps/task/evol_instruct/evol_complexity/utils.py
```diff
@@ -12,35 +12,36 @@
...
-    )
+GENERATION_MUTATION_TEMPLATES = {
+    "FRESH_START": GENERATION_MUTATION_TEMPLATES_EVOL_INSTRUCT["FRESH_START"],
+    "CONSTRAINTS": GENERATION_MUTATION_TEMPLATES_EVOL_INSTRUCT["CONSTRAINTS"],
+    "DEEPENING": GENERATION_MUTATION_TEMPLATES_EVOL_INSTRUCT["DEEPENING"],
+    "CONCRETIZING": GENERATION_MUTATION_TEMPLATES_EVOL_INSTRUCT["CONCRETIZING"],
+    "INCREASED_REASONING_STEPS": GENERATION_MUTATION_TEMPLATES_EVOL_INSTRUCT[
+        "INCREASED_REASONING_STEPS"
+    ],
+}
```
gabrielmbmb, line 39: maybe we can just simply `GENERATION_MUTATION_TEMPLATES = GENERATION_MUTATION_TEMPLATES_EVOL_INSTRUCT.copy()`

gabrielmbmb (approved):

src/distilabel/steps/task/evol_instruct/evol_complexity/utils.py
plaguss: They don't share the "BREADTH", we can let like this or copy and remove that field

gabrielmbmb: oh my bad

PR #483, opened 2024-03-26
Update `_WriteBuffer` to write several parquet files

gabrielmbmb:
## Description

This PR updates the `_WriteBuffer` to create several parquet files instead of just one using the `ParquetWriter`. As we're inferring the schema using the first row added to the `_WriteBuffer`, we needed to check that schema for each batch is equal to the writer schema, and if it wasn't, create a new one unifying both schemas. This required closing the `ParquetWriter` and reopening it with the new schema, but this overwrites the content of the parquet file, losing the data from previous steps.

With this new approach, we write a parquet file (`00001.parquet`, `00002.parquet`, etc) once the buffer for the step is full. The buffer will be considered to be full once the defined number of elements in the buffer is reached.

### DIFF
--- a/src/distilabel/pipeline/base.py
+++ b/src/distilabel/pipeline/base.py
@@ -34,6 +34,7 @@ from typing_extensions import Self
 
 from distilabel import __version__
 from distilabel.pipeline._dag import DAG
+from distilabel.utils.files import list_files_in_dir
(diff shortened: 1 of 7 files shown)


plaguss (approved):
lgtm

PR #489, opened 2024-03-28
Use `QueueHandler` for `Pipeline` logging

gabrielmbmb:
## Description

This PR updates the logging setup for the local `Pipeline` to use a `QueueHandler` and a `QueueListener` (a thread in the main process). With this setup, child processes send log messages using a `multiprocessing.Queue` and the `QueueHandler` to the `QueueListener`, which will be in charge of handling the log messages.

This change was mainly motivated to improve the logging messages in environments like Google Colab and IPython Notebooks, in which not all the logging messages (from the child processes) were being displayed.

image

### DIFF
--- a/src/distilabel/llm/base.py
+++ b/src/distilabel/llm/base.py
@@ -70,8 +70,8 @@ class LLM(RuntimeParametersMixin, BaseModel, _Serializable, ABC):
         per input in `inputs`."""
         pass
 
-    @cached_property
-    def generate_parameters(self) -> List[inspect.Parameter]:
+    @property
+    def generate_parameters(self) -> List["inspect.Parameter"]:
         """Returns the parameters of the `generate` method.
 
         Returns:
--- a/src/distilabel/llm/base.py
+++ b/src/distilabel/llm/base.py
@@ -27,7 +27,6 @@ from distilabel.mixins.runtime_parameters import (
     RuntimeParametersMixin,
 )
 from distilabel.utils.docstring import parse_google_docstring
-from distilabel.utils.logging import get_logger
 from distilabel.utils.serialization import _Serializable
 
 if TYPE_CHECKING:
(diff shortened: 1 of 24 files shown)


src/distilabel/llm/base.py
```diff
@@ -72,8 +70,8 @@ def generate(
         per input in `inputs`."""
         pass
 
-    @cached_property
-    def generate_parameters(self) -> List[inspect.Parameter]:
+    @property
```
alvarobartt, line 73: Then the `from functools import cached_property` can be removed too, right?

src/distilabel/steps/task/generate_embeddings.py
```diff
@@ -45,6 +45,8 @@ class GenerateEmbeddings(Step):
     llm: LLM
 
     def load(self) -> None:
+        """Loads the `LLM` used to generate the embeddings."""
+        super().load()
```
alvarobartt, line 49: If all the steps inheriting from `Step` need to actually call the `super().load()` isn't it better if we just add it as a separate method that runs within the `post_init`? i.e. `model_post_init` on `_Step`?

alvarobartt (approved):

src/distilabel/steps/task/generate_embeddings.py
gabrielmbmb: Sadly we can't do that :( We need to create the logger after having called `setup_logging` in each process.

src/distilabel/llm/base.py
gabrielmbmb: yes! BTW I removed it because in some environments `pickle` was not able to serialize it.

src/distilabel/steps/task/generate_embeddings.py
alvarobartt: Can we then just do `step.load()` and then `step.setup_logging`? But to manage that within the `Pipeline` instead of having to always add that as part of the `load` method :/

PR #494, opened 2024-04-01
Draw Pipeline DAG using graphviz

plaguss:
## Description

This PR adds a mixin class to draw the `DAG` of the pipelines.

Also, it includes `RenameColumns` step as it is useful enough as to keep it as a module.

An example for the following pipeline:

```python
from distilabel.steps.generators.huggingface import LoadHubDataset
from distilabel.steps.combine import CombineColumns
from distilabel.steps.rename import RenameColumns
from distilabel.steps.task.evol_instruct.base import EvolInstruct
from distilabel.llm.huggingface.inference_endpoints import InferenceEndpointsLLM
from distilabel.steps.task.pair_rm import PairRM


if __name__ == "__main__":
    with Pipeline(name="PairRM pipe") as pipeline:
        load_hub_dataset = LoadHubDataset(name="load_dataset", batch_size=8)
        rename_columns = RenameColumns(
            name="rename_columns",
            rename_mappings={"prompt": "instruction"},
            input_batch_size=12,
        )
        evol_instruction_complexity_1 = EvolInstruct(
            name="evol_complexity_openai",
            llm=OpenAILLM(model="gpt-3.5-turbo"),
            num_evolutions=5,
            store_evolutions=True,
            generate_answers=True,
            include_original_instruction=True,
        )
        evol_instruction_complexity_2 = EvolInstruct(
            name="evol_complexity_huggingface",
            llm=InferenceEndpointsLLM(model_id="mistralai/Mixtral-8x7B-Instruct-v0.1"),
            num_evolutions=5,
            store_evolutions=True,
            generate_answers=True,
            include_original_instruction=True,
        )
    
        load_hub_dataset.connect(rename_columns)
        rename_columns.connect(evol_instruction_complexity_1)
        rename_columns.connect(evol_instruction_complexity_2)

        # Will combine the outputs from the different LLMs (the candidates to rank)
        combine_columns = CombineColumns(
            name="combine_columns",
            columns=["response"],
            output_columns=["candidates"],
        )
        
        evol_instruction_complexity_1.connect(combine_columns)
        evol_instruction_complexity_2.connect(combine_columns)

        # Prepares the column names to the name expected from `PairRM`
        rename_for_pairrm = RenameColumns(
            name="rename_for_pairrm",
            input_batch_size=12,
        )
        combine_columns.connect(rename_for_pairrm)

        ranker = PairRM(name="pair_rm_ranker")
        rename_for_pairrm.connect(ranker)

    pipeline.render(view=True)
```

Can be seen at the following figure:
PairRM_pipe

There are certain "style" decisions that can be discussed if we want to include this:
- Currently it uses a different (random) color depending on the type of step (`GeneratorStep`, `GlobalStep`, `Task` or plain `Step`).
- Certain number of arguments are removed to simplify visualization of the nodes (steps) in the figure.
- Arguments are capped to 30 characters to keep the nodes relatively small.
- The figure could be optionally included when we push a `Distiset` to the hub, so it can be shown in the dataset card.
- The figure is saved in the same cache directory by default.

### DIFF
--- a/src/distilabel/pipeline/base.py
+++ b/src/distilabel/pipeline/base.py
@@ -34,6 +34,7 @@ import pyarrow.parquet as pq
 from typing_extensions import Self
 
 from distilabel import __version__
+from distilabel.mixins.graph_viz import GraphvizMixin
 from distilabel.pipeline._dag import DAG
 from distilabel.utils.files import list_files_in_dir
 from distilabel.utils.serialization import TYPE_INFO_KEY, _Serializable
@@ -85,7 +86,7 @@ class _GlobalPipelineManager:
         return cls._context_global_pipeline
 
 
-class BasePipeline(_Serializable):
+class BasePipeline(GraphvizMixin, _Serializable):
     """Base class for a `distilabel` pipeline.
 
     Attributes:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,9 +45,9 @@ jobs:
           python_version=$(python -c "import sys; print(sys.version_info[:2])")
 
           if [ "${python_version}" != "(3, 8)" ]; then
-            pip install -e .[dev,tests,hf-transformers,hf-inference-endpoints,vertexai,ollama,openai,together,argilla,mistralai,llama-cpp,anthropic,litellm]
+            pip install -e .[dev,tests,hf-transformers,hf-inference-endpoints,vertexai,ollama,openai,together,argilla,mistralai,llama-cpp,anthropic,litellm,graphviz]
           else
-            pip install -e .[dev,tests,hf-transformers,hf-inference-endpoints,vertexai,ollama,openai,together,argilla,llama-cpp,anthropic,litellm]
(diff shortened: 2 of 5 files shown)


src/distilabel/pipeline/base.py
```diff
@@ -34,6 +34,7 @@
 from typing_extensions import Self
 
 from distilabel import __version__
+from distilabel.mixins.graph_viz import GraphvizMixin
```
alvarobartt, line 37: Even though `graphviz` has been defined as optional, when importing it here we make it mandatory and the user will need to install it even if not planning to use it, so maybe then we can make the dependency mandatory? Otherwise we can just define a function to do so instead of a `MixIn`?

alvarobartt (commented):
Looks good! Just one minor comment regarding the dependency 👍🏻

src/distilabel/pipeline/base.py
plaguss: Totally! Will take a look, maybe it's better to avoid adding the dependency, so it can be kept in a function for simplicity.

PR #498, opened 2024-04-02
Added authentication header to allow private/gated dataset use

bjoernpl:
Before, I would receive a `Failed to get '{repo_id}' dataset info.` error when attempting to load a gated dataset. Now, this uses the `HF_TOKEN` env variable.

Probably there's a more principled way that would use the `HF_TOKEN` saved in the `HF_HOME` directory after calling `huggingface-cli login` if the `HF_TOKEN` env variable is not set.

### DIFF
--- a/src/distilabel/cli/pipeline/utils.py
+++ b/src/distilabel/cli/pipeline/utils.py
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, List, Tuple
-import os
 
 import requests
 import yaml
--- a/src/distilabel/cli/pipeline/utils.py
+++ b/src/distilabel/cli/pipeline/utils.py
@@ -14,6 +14,7 @@
 
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, List, Tuple
+import os
 
 import requests
 import yaml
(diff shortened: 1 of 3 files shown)


alvarobartt (approved):
Hi here! Thanks for the PR, LGTM! Only one nit, could you also add that to the `argilla` integration within `distilabel`? Mainly because Argilla can also be deployed in Spaces and we also need to provide the auth headers if it's behind a private Space 👍🏻 If you don't have time or prefer me to do it, just let me know, otherwise I'm happy to merge!

bjoernpl:
Sure, if you can point me to the places where this interaction happens, I can have a look.

alvarobartt:
Sure, it should be the following lines:

https://github.com/argilla-io/distilabel/blob/251e34c3c6f571feaa92cc7f38407295118850de/src/distilabel/steps/argilla/base.py#L79-L84 

And those should be included within the `rg.init` call as described in the Argilla documentation at https://docs.argilla.io/en/latest/getting_started/installation/deployments/huggingface-spaces.html#create-your-first-dataset

`rg.init(api_url="<ARGILLA_API_URL>", api_key="<ARGILLA_API_KEY>", extra_headers={"Authorization": f"Bearer {os.environ['HF_TOKEN']}"})`

And the URL looks like `https://alvarobartt-my-argilla.hf.space/datasets/`, so I think it's safe to say that if `self.base_url.__contains__("hf.space"): ...` then add the extra headers if the `HF_TOKEN` is provided 👍🏻

bjoernpl:
Maybe you can test if this works? Don't have the workflow for this set up yet.

PR #501, opened 2024-04-03
Fix `asyncio` in `AsyncLLM` to use the running event loop if any

alvarobartt:
## Description

This PR adds a hot-fix for the `AsyncLLM` class and subclasses, since the `event_loop` is always created, but in Colab there's already an active running event loop, so that should be reused, otherwise the `AsyncLLM` subclasses or any other `asyncio` code creating `event_loop`s will fail in Colab with the following exception `RuntimeError: Cannot run the event loop while another loop is running`.

On top of that, an utils function named `in_notebook` has been included so as to identify whether the code is running in a notebook, and if so, use `nest-asyncio` so as to be able to successfully use `asyncio`.

Thanks @burtenshaw for reporting and @gabrielmbmb for the hints!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ dependencies = [
     "importlib-resources >= 6.1.1; python_version < '3.9'",
     "Jinja2 >= 3.1.2",
     "multiprocess >= 0.70",
+    "nest-asyncio >= 1.6.0",
     "networkx >= 3.0",
     "pydantic >= 2.0",
     "rich >= 13.5.0",
--- a/src/distilabel/llm/base.py
+++ b/src/distilabel/llm/base.py
@@ -27,7 +27,7 @@ from distilabel.mixins.runtime_parameters import (
     RuntimeParametersMixin,
 )
 from distilabel.utils.docstring import parse_google_docstring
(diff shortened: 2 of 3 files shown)


gabrielmbmb (approved):
LGTM!

burtenshaw:
How did you test this? I merged it into the DEITA branch `docs/deita-tutorial` and tried it in the notebook but I got the same error.

gabrielmbmb:
For me it works, but I needed to use `nest_asyncio`:

```python
import nest_asyncio
from distilabel.steps.task.evol_instruct.base import EvolInstruct

nest_asyncio.apply()

pipeline = Pipeline(name="DEITA")

evol_instruction_complexity = EvolInstruct(
    name="evol_instruction_complexity",
    llm=OpenAILLM(model="gpt-3.5-turbo"),
    num_evolutions=5,
    store_evolutions=True,
    generate_answers=True,
    include_original_instruction=True,
    pipeline=pipeline,
)

next(evol_instruction_complexity.process(([{"instruction": "How many fish are there in a dozen fish?"}])))
```

maybe we can add `nest_asyncio` as dependency and in `distilabel/__init__.py` check if running on ipython environment, and if so `nest_asyncio.apply()`

PR #502, opened 2024-04-03
Docs cli

plaguss:
## Description

This PR adds a small guide to use the CLI.


davidberenstein1957 (commented):
maybe we can add redirects to the API reference pages in the docs?

plaguss:
> maybe we can add redirects to the API reference pages in the docs?

I added a section in the API reference for the CLI which didn't exist, but in this case we have more information in the tutorial than in the reference itself, as the code is internally built by typer for the commands, and it's better expressed via the `--help` of the commands

PR #508, opened 2024-04-08
Add `CohereLLM`

gabrielmbmb:
## Description

Add `CohereLLM` class which allows to use Cohere Chat API for generating text.

## Example

```python
from distilabel.llms import CohereLLM

llm = CohereLLM(model="command-r", api_key="...")
llm.load()

results = llm.generate(
    inputs=[
        [
            {
                "role": "system",
                "content": "You're are a pirate. You only generate text that sounds like a pirate.",
            },
            {
                "role": "user",
                "content": "How many helicopters can human eat in one sitting?",
            },
            {
                "role": "assistant",
                "content": "Arggg, that's not possible scallywag!",
            },
            {
                "role": "user",
                "content": "and if it's a mutant human?",
            },
        ]
    ],
    num_generations=3,
)
```


alvarobartt:
🚨 API Key leak @gabrielmbmb!

PR #509, opened 2024-04-08
Add `docs/sections/learn/argilla.md`

alvarobartt:
## Description

This PR adds the `docs/sections/learn/argilla.md` section with a brief explanation on why Argilla is important, and the different steps that have been developed around Argilla, and how to easily integrate those within a working `Pipeline` with one working example per task.

Besides that, also the formatting around some `#` characters has been fixed.


plaguss (commented):
Really nice! I would also add an autocontained example (maybe as a dropdown?) of how to set up argilla, maybe just using the hub space, so the user doesn't have to take a look at how to do it in the argilla docs).

gabrielmbmb (approved):

PR #522, opened 2024-04-11
pass on Distiset's kwargs to Dataset.push_to_hub()

rasdani:
I want to specify a `revision` for my datasets. 

Therefor I'm passing on the kwargs through the `Distiset` wrapper to HF `Dataset`.

Didn't get around to testing today, though.

### DIFF
--- a/src/distilabel/distiset.py
+++ b/src/distilabel/distiset.py
@@ -46,7 +46,7 @@ class Distiset(dict):
         private: bool = False,
         token: Optional[str] = None,
         generate_card: bool = True,
-        **kwargs,
+        **kwargs: Any,
     ) -> None:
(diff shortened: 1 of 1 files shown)


src/distilabel/distiset.py
```diff
@@ -46,6 +46,7 @@ def push_to_hub(
         private: bool = False,
         token: Optional[str] = None,
         generate_card: bool = True,
+        **kwargs,
```
gabrielmbmb, line 49: can we add `Any` type hint to `**kwargs` and add also the docstring?

gabrielmbmb (changes requested):
Hi @rasdani, LGTM and thanks! 

just a small change required before merging

PR #530, opened 2024-04-15
Fix error in README.md when pushing the custom dataset card

plaguss:
## Description

We have a bug when calling `Distiset.push_to_hub(generate_card=True)` as the metadata is overwritten. This PR fixes it by downloading the current README.md and updating the information.

### DIFF
--- a/src/distilabel/distiset.py
+++ b/src/distilabel/distiset.py
@@ -15,7 +15,7 @@
 import logging
 import re
 from pathlib import Path
-from typing import Optional
+from typing import Any, Dict, Optional
 
 from datasets import load_dataset
 from huggingface_hub import DatasetCardData, HfApi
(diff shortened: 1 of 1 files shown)


src/distilabel/distiset.py
```diff
@@ -91,13 +91,16 @@ def _generate_card(self, repo_id: str, token: Optional[str]) -> None:
                 dataset[0] if not isinstance(dataset, dict) else dataset["train"][0]
             )
 
+        metadata = self._extract_readme_metadata(repo_id, token)
+
         card = DistilabelDatasetCard.from_template(
             card_data=DatasetCardData(
                 config_names=sorted(self.keys()),
```
gabrielmbmb, line 98:
```suggestion
```

I think we can remove this line.

gabrielmbmb (approved):
LGTM! Just a small suggestion.

PR #537, opened 2024-04-15
Fix parsing `LLM` generation kwargs

gabrielmbmb:
## Description

This PR fixes an issue for which `LLM.generate` or `LLM.agenerate` generation kwargs (which are `RuntimeParameter`s) were not being coerced to its type. For example, if `max_new_tokens="512"` then it won't get coerced to `int` type causing an error later. To avoid this issue, `generate` and `agenerate` methods of the `LLM`s must be decorated with `pydantic.validate_call`, which coerces the provides values to the arguments type.

This was a problem specially when specifying the values for the generation kwargs of the LLM using the CLI, as from the CLI everything is a string.

### DIFF
--- a/docs/sections/learn/llms/index.md
+++ b/docs/sections/learn/llms/index.md
@@ -110,7 +110,7 @@ class CustomLLM(LLM):
         return "my-model"
 
     @validate_call
-    def generate(self, inputs: List[ChatType], num_generations: int = 1) -> List[GenerateOutput]:
+    def generate(self, inputs: List[ChatType], num_generations: int = 1, **kwargs: Any) -> List[GenerateOutput]:
         for _ in range(num_generations):
             ...
 
--- a/docs/sections/learn/llms/index.md
+++ b/docs/sections/learn/llms/index.md
@@ -98,6 +98,8 @@ Once those methods have been implemented, then the custom LLM will be ready to b
 ```python
 from typing import Any
 
+from pydantic import validate_call
+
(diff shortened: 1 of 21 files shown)


src/distilabel/llms/ollama.py
```diff
@@ -12,17 +12,56 @@
...
+from distilabel.steps.tasks.typing import ChatType
 
 if TYPE_CHECKING:
-    from ollama import AsyncClient, Options
-
-    from distilabel.steps.tasks.typing import ChatType
+    from ollama import AsyncClient
+
+
+class Options(TypedDict, total=False):
```
alvarobartt, line 28: Maybe you can add the reference here to the original file within `ollama`?

docs/sections/learn/llms/index.md
```diff
@@ -107,6 +109,7 @@ class CustomLLM(LLM):
     def model_name(self) -> str:
         return "my-model"
 
+    @validate_call
     def generate(self, inputs: List[ChatType], num_generations: int = 1) -> List[GenerateOutput]:
```
alvarobartt, line 113:
```suggestion
    def generate(self, inputs: List[ChatType], num_generations: int = 1, **kwargs: Any) -> List[GenerateOutput]:
```

alvarobartt (commented):
LGTM, but maybe we should also add that within the docs so that users use the `validate_call` decorator on the classes they subclass? i.e. to mention that in the custom step/task/llm sections, thanks!

PR #539, opened 2024-04-16
docs: update documentation for huggingface inference endpoints.

burtenshaw:
This PR updates the documentation for the `InferenceEndpointsLLM` and expands for the main uses: free, dedicated, TGI

### DIFF
--- a/src/distilabel/llms/huggingface/inference_endpoints.py
+++ b/src/distilabel/llms/huggingface/inference_endpoints.py
@@ -59,10 +59,35 @@ class InferenceEndpointsLLM(AsyncLLM):
         use_openai_client: whether to use the OpenAI client instead of the Hugging Face client.
 
     Examples:
-        >>> from distilabel.llms.huggingface import AsyncInferenceEndpointsLLM
-        >>> llm = AsyncInferenceEndpointsLLM(model_id="model-id")
-        >>> llm.load()
-        >>> output = await llm.agenerate([{"role": "user", "content": "Hello world!"}])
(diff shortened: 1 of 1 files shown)


src/distilabel/llms/huggingface/inference_endpoints.py
```diff
@@ -59,10 +59,35 @@ class InferenceEndpointsLLM(AsyncLLM):
         use_openai_client: whether to use the OpenAI client instead of the Hugging Face client.
 
     Examples:
-        >>> from distilabel.llms.huggingface import AsyncInferenceEndpointsLLM
-        >>> llm = AsyncInferenceEndpointsLLM(model_id="model-id")
-        >>> llm.load()
-        >>> output = await llm.agenerate([{"role": "user", "content": "Hello world!"}])
+        from distilabel.llms.huggingface import InferenceEndpointsLLM
```
gabrielmbmb, line 62: we need to either put this between a markdown code block or to use `>>>` again to get it render in the UI

image

gabrielmbmb (changes requested):

src/distilabel/llms/huggingface/inference_endpoints.py
alvarobartt: As well as adding the code-comments as `>>> # ...` as @gabrielmbmb mentioned above

burtenshaw: I went with fences because they support the copy button.

PR #540, opened 2024-04-16
Set `config="default"` in `Distiset` when only one leaf `Step`

alvarobartt:
## Description

This PR sets the configuration name in the `Distiset` to default if there's only one leaf `Step`, so that the dataset can be simply loaded as `load_dataset("dataset-name")` without having to provide the configuration, as that's more aligned with the current configuration handling within the Hugging Face Hub.

Additionally, this PR fixes some typos within the `docs` as identified with `codespell`, also aligns the naming for `Hugging Face` to use that as a standard over `huggingface`, `HuggingFace`, `Huggingface`, and such; and also removes some unused files within the docs that were copied over during the merge with `main` from `core-refactor`, recently.


src/distilabel/pipeline/base.py
```diff
@@ -189,7 +188,7 @@ def run(
         self,
         parameters: Optional[Dict[str, Dict[str, Any]]] = None,
         use_cache: bool = True,
-    ) -> "Distiset":  # type: ignore
+    ) -> None:
```
gabrielmbmb, line 191: This was like this on purpose, to denote that classes extending this method must return a `Distiset`.

gabrielmbmb (changes requested):
LGTM! Just a comment with the updated type hints of `run` method.

src/distilabel/pipeline/base.py
alvarobartt: Hmm then shouldn't that be a protected / private method as that's just called via `super()` and instead create an `abstractmethod` for the `run` method?

alvarobartt: I can revert it for the moment, but that felt off to me :/

gabrielmbmb: We can create a `_run` abstractmethod that can be later be invoked by `run`, but I think that's adding complexity which is not needed. The docstring also specified that this method is meant to be extended.

gabrielmbmb (approved):

PR #558, opened 2024-04-23
Split `ChatGeneration` from `TextGeneration`

alvarobartt:
## Description

This PR splits the `ChatGeneration` task from the `TextGeneration` one, since the naming was confusing since both cases when:

* A instruction was provided
* A list of messages / conversation was provided

were using the `instruction` name, while using a different `format_input` function underneath.

So on, this PR splits that from `TextGeneration` while keeping it for `instruction` only, but raising a `DeprecationWarning` when using that with a conversation, to be deprecated in v1.2.0; while moving that functionality to `ChatGeneration` under the input `messages` instead.

Additionally, this PR also adds the `system_prompt` as an optional input so that it's included within the `TextGeneration` task if existing within the input batch and if the flag `use_system_prompt` is set to `True`.

## Example

```python
import os

from distilabel.llms.mistral import MistralLLM
from distilabel.steps.tasks.text_generation import TextGeneration

text_generation = TextGeneration(
    name="chat-generation",
    llm=MistralLLM(
        model="mistral-tiny",
        api_key=os.getenv("MISTRALAI_API_KEY"),  # type: ignore
    ),
    use_system_prompt=True,
    input_batch_size=8,
    pipeline=Pipeline(name="text-gen-pipeline"),
)
text_generation.load()

result = next(
    text_generation.process(
        [
            {
                "instruction": "What if the Beatles had never formed as a band?",
                "system_prompt": "You are a helpful assistant, that knows everything about The Beatles."
            },
        ] 
    )
)
```


tests/unit/steps/tasks/test_generation.py
gabrielmbmb, line 1: Can you add tests for checking that `format_input` raises `ValueError`s? I did not, sorry

gabrielmbmb (changes requested):
LGTM! We need to update this section in the docs: https://distilabel.argilla.io/latest/sections/learn/tasks/text_generation/

src/distilabel/steps/tasks/generation.py
```diff
@@ -0,0 +1,110 @@
...
+                stacklevel=2,
+            )
+            return input["instruction"]
+
+        if not isinstance(input["instruction"], str):
+            raise ValueError(
+                f"Input `instruction` must be a string. Got: {input['instruction']}."
+            )
+
+        return [{"role": "user", "content": input["instruction"]}]
```
gabrielmbmb, line 60: With this, the system prompt can no longer be defined.

src/distilabel/steps/tasks/generation.py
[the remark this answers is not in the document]
alvarobartt: But that's misleading somehow, right?

src/distilabel/steps/tasks/generation.py
alvarobartt: Fixing that!

gabrielmbmb: It would be simpler just to have `TextGeneration` and `messages` as input. If you want to do `response` generation for an `instruction`, then just pass:

```python
[
    {"role": "system", "content": "..."},
    {"role": "user", "content": "..."},
]
```

Simpler to maintain and less confusing.

alvarobartt: But that would imply extra pre-processing steps since most of the datasets contain raw prompts, not OpenAI formatted prompts, and I think the idea is to make it easier to use? Also this should be also easy to maintain right? It's a task inheriting from the base `TextGeneration`, WDYT @plaguss @gabrielmbmb?

gabrielmbmb (approved):

gabrielmbmb (approved):

PR #564, opened 2024-04-24
Fix `RuntimeParamater` validation when provided as `_Step` attr

### LINKED ISSUE #563: [BUG] `model_path` in `LlamaCppLLM` is superseded by `RuntimeParameter`
**Describe the bug**

The `model_path` in `LlamaCppLLM` cannot be set via attribute, since it will be superseded with the `RuntimeParameter` which defaults to `None`, meaning that `LlamaCppLLM` cannot be used.

Since even when providing the `model_path` via runtime parameter, it won't work, and the pipeline will trigger the same error, so `LlamaCppLLM` cannot be used from `distilabel` v1.0.0 onwards.

**To Reproduce**

```python
from distilabel.llms import LlamaCppLLM
from distilabel.steps import LoadHubDataset
from distilabel.steps.tasks import TextGeneration

if __name__ == "__main__":
    with Pipeline(name="example") as pipeline:
            name="load_dataset",
            output_mappings={"prompt": "instruction"},
        )

        text_generation = TextGeneration(
            name="text_generation",
            llm=LlamaCppLLM(
                model_path="",  # !wget https://huggingface.co/TheBloke/zephyr-7B-beta-GGUF/blob/main/zephyr-7b-beta.Q4_K_M.gguf
                n_gpu_layers=-1,
                verbose=True,
            ),
            input_batch_size=10,
            output_mappings={"model_name": "generation_model"},
        )
        load_hub_dataset.connect(text_generation)
```

**Expected behaviour**

The `RuntimeParameter` check to be skipped since it's already provided within the `LlamaCppLLM` class instance.

**Desktop:**
 - Package version: latest from `develop`
 - Python version: 3.11

alvarobartt:
## Description

This PR fixes a bug within the `RuntimeParameter` validation, since it was failing even when the mandatory values where provided via `_Step` attributes.

This way, the mandatory `RuntimeParameters` are ensured to be not `None`, while the optional ones can indeed be `None`.

For traceability the bug was introduced in https://github.com/argilla-io/distilabel/pull/493, but related to https://github.com/argilla-io/distilabel/pull/466

Closes #563

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
           else
             pip install -e .[dev,tests,cohere,hf-transformers,hf-inference-endpoints,vertexai,ollama,openai,together,argilla,llama-cpp,anthropic,litellm]
           fi
-          pip install git+https://github.com/yuchenlin/LLM-Blender.git@3c2d71f
+          pip install git+https://github.com/argilla-io/LLM-Blender.git
 
       - name: Lint
         run: make lint
--- a/src/distilabel/__init__.py
+++ b/src/distilabel/__init__.py
@@ -14,6 +14,6 @@
 
 from rich import traceback as rich_traceback
 
-__version__ = "1.0.1"
+__version__ = "1.0.2"
 
 rich_traceback.install(show_locals=True)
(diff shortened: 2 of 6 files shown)


gabrielmbmb (approved):
LGTM!

PR #565, opened 2024-04-24
Set `streaming=False` and add `num_examples` to `LoadHubDataset`

### LINKED ISSUE #561: [FEATURE] Update `LoadHubDataset` so it can work offline
The `LoadHubDataset` step can only work if connected to Hugging Face.

- Remove the hardcoded `streaming=True` parameter from `load_dataset` and instead transform it to a `RuntimeParameter`
- By default do not use `streaming`
- If dataset has been downloaded, then try to get the name of the columns from the locally cached dataset instead of calling `datasets-server.huggingface.co` API.

plaguss:
## Description

This PR modifies the default behaviour of `LoadHubDataset` to use `streaming=False` as default, and tries to fetch the column info from the cached dataset if found.

Closes #561, and also includes the functionality from PR #520:

```python
from distilabel.steps.generators.huggingface import LoadHubDataset

    name="load_dataset",
    repo_id="HuggingFaceH4/instruction-dataset",
    split="test",
    batch_size=8,
    num_examples=4,
    pipeline=Pipeline(name="dataset-pipeline"),
)
```

### DIFF
--- a/src/distilabel/steps/generators/huggingface.py
+++ b/src/distilabel/steps/generators/huggingface.py
@@ -84,6 +84,9 @@ class LoadHubDataset(GeneratorStep):
         - `split`: The split of the dataset to load. Defaults to 'train'.
         - `config`: The configuration of the dataset to load. This is optional and only
             needed if the dataset has multiple configurations.
+        - `streaming`: Whether to load the dataset in streaming mode or not. Defaults to False.
+        - `num_examples`: The number of examples to load from the dataset.
+            By default will load all examples.
 
     Output columns
         - dynamic, based on the dataset being loaded
@@ -106,6 +109,10 @@ class LoadHubDataset(GeneratorStep):
         default=False,
         description="Whether to load the dataset in streaming mode or not. Defaults to False.",
     )
+    num_examples: RuntimeParameter[int] = Field(
+        default=None,
+        description="The number of examples to load from the dataset. By default will load all examples.",
+    )
 
     _dataset: Union[IterableDataset, None] = PrivateAttr(...)
 
@@ -119,6 +126,13 @@ class LoadHubDataset(GeneratorStep):
             split=self.split,
             streaming=self.streaming,
         )
+        num_examples = self._get_dataset_num_examples()
+        self.num_examples = (
+            min(self.num_examples, num_examples) if self.num_examples else num_examples
+        )
+
+        if not self.streaming:
+            self._dataset = self._dataset.select(range(self.num_examples))
 
     def process(self, offset: int = 0) -> "GeneratorStepOutput":
         """Yields batches from the loaded dataset from the Hugging Face Hub.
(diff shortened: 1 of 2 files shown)


src/distilabel/steps/generators/huggingface.py
```diff
@@ -136,7 +154,7 @@ def process(self, offset: int = 0) -> "GeneratorStepOutput":
             transformed_batch = self._transform_batch(batch)
             batch_size = len(transformed_batch)
             num_returned_rows += batch_size
-            yield transformed_batch, num_returned_rows == num_examples
+            yield transformed_batch, num_returned_rows == self.num_examples
```
alvarobartt, line 157: Maybe I'm missing something, but shouldn't this be greater or equal instead?

```suggestion
            yield transformed_batch, num_returned_rows >= self.num_examples
```

src/distilabel/steps/generators/huggingface.py
```diff
@@ -136,7 +154,7 @@ def process(self, offset: int = 0) -> "GeneratorStepOutput":
             transformed_batch = self._transform_batch(batch)
             batch_size = len(transformed_batch)
             num_returned_rows += batch_size
```
alvarobartt, line 156: If the comment below is the case i.e. we can return more data that the specified data, shouldn't we count all the total rows, and on each iter `if num_returned_rows > self.num_examples` then `transformed_batch[:(self.num_examples - num_returned_rows)]`?

Meaning that if we have say 10 rows and we use batches of size 5, but `num_examples` is set to 9, then we batch that so that we return 5 and then 5 too, since we're not checking that the total number of rows is bigger than `num_examples` right?

tests/unit/steps/generators/test_huggingface.py
```diff
@@ -0,0 +1,65 @@
...
+        assert isinstance(load_hub_dataset._dataset, ds_type)
+
+        generator_step_output = next(load_hub_dataset.process())
+        assert isinstance(generator_step_output, tuple)
+        assert isinstance(generator_step_output[1], bool)
+        assert len(generator_step_output[0]) == 2
+
+    def test_dataset_outputs(self, dataset_loader: LoadHubDataset) -> None:
+        # TODO: This test can be run with/without internet connection, we should emulate it here with a mock.
+        assert dataset_loader.outputs == ["prompt", "completion", "meta"]
```
alvarobartt, line 65: Maybe we can just add an integration test for the main use cases of this feature? i.e. downloading a dataset from the Hub and then ensuring that's not downloaded again.

alvarobartt (commented):
LGTM, just some minor comments about `num_examples` and whether that can introduce misalignments or not, thanks!

src/distilabel/steps/generators/huggingface.py
plaguss: I changed the condition below because it makes more sense to have `>=` there, we don't have the problem anymore because once filtered the dataset even if we wanted to retrieve more rows, the dataset already has 9 examples instead of 10, so we won't yield more

alvarobartt (approved):

PR #566, opened 2024-04-24
Make `pipeline` argument of `Step` optional

### LINKED ISSUE #554: [FEATURE] Make `pipeline` argument of `Step` optional
## Description

When creating an `Step` it's required to pass a `pipeline` argument or to create the `Step` within the pipeline context manager, so one is assigned automatically. This can be cumbersome when creating an instance of an `Step` just to test it.

I propose modifying the `model_post_init` function of the `_Step` class to give a warning to the user instead of raising an error when `pipeline==None`.

plaguss:
## Description

Make the `_Step.model_post_init` less strict to allow instantiating steps without a `Pipeline` and throwing a warning instead of raising a `ValueError`.

This should simplify testing steps without the overhead of passing a `Pipeline` (but the error will propagate as the warning indicates if the `Pipeline` is needed while testing).

Closes #554.


src/distilabel/steps/base.py
```diff
@@ -113,13 +113,18 @@ def model_post_init(self, __context: Any) -> None:
             self.pipeline = _GlobalPipelineManager.get_pipeline()
 
         if self.pipeline is None:
-            raise ValueError(
+            if not isinstance(self._logger, logging.Logger):
+                _logger = logging.getLogger(f"distilabel.step.{self.name}")
+            else:
+                _logger = self._logger
+            _logger.warning(
```
alvarobartt, line 120: What's the reason behind this change? Because AFAIK if the `self._logger` is initialized within the `load` call, it's never going to be something different than `logging.Logger`, right? i.e. the only scenario is whenever `self.load` is not called, right? cc @gabrielmbmb to confirm

plaguss: I've updated this to always get a logger at this stage (it wasn't throwing errors, but it was wrong anyway):
```python
            _logger = logging.getLogger(f"distilabel.step.{self.name}")
            _logger.warning(
                f"Step '{self.name}' hasn't received a pipeline, and it hasn't been"
                " created within a `Pipeline` context. Please, use"
                " `with Pipeline() as pipeline:` and create the step within the context."
            )
```
When this method runs we don't have yet the logger, and we want something to inform the user of the fact, but maybe the load method does something different than just loading the logger, so I think it's a good compromise, WDYT?

tests/unit/pipeline/test_dag.py
```diff
@@ -504,15 +504,15 @@ def test_dag_from_dict(self, dummy_step_1: "Step", dummy_step_2: "Step") -> None
         assert "dummy_step_2" in new_dag.G["dummy_step_1"]
 
     def test_dag_from_dict_errored_without_pipeline(
-        self, dummy_step_1: "Step", dummy_step_2: "Step"
+        self, caplog, dummy_step_1: "Step", dummy_step_2: "Step"
```
alvarobartt, line 507: Maybe it's nice to also add the type-hint here

```suggestion
        self, caplog: pytest.LogCaptureFixture, dummy_step_1: "Step", dummy_step_2: "Step"
```

tests/unit/steps/tasks/evol_instruct/test_base.py
```diff
@@ -45,14 +45,14 @@ def test_within_pipeline_context(self, dummy_llm: LLM) -> None:
             assert task.llm is dummy_llm
         assert task.pipeline == pipeline
 
-    def test_with_errors(self, dummy_llm: LLM) -> None:
+    def test_with_errors(self, caplog, dummy_llm: LLM) -> None:
```
alvarobartt, line 48:
```suggestion
    def test_with_errors(self, caplog: pytest.LogCaptureFixture, dummy_llm: LLM) -> None:
```

tests/unit/steps/tasks/evol_instruct/test_generator.py
```diff
@@ -45,16 +45,16 @@ def test_within_pipeline_context(self, dummy_llm: LLM) -> None:
             assert task.llm is dummy_llm
         assert task.pipeline == pipeline
 
-    def test_with_errors(self, dummy_llm: LLM) -> None:
+    def test_with_errors(self, caplog, dummy_llm: LLM) -> None:
```
alvarobartt, line 48:
```suggestion
    def test_with_errors(self, caplog: pytest.LogCaptureFixture, dummy_llm: LLM) -> None:
```

tests/unit/steps/tasks/evol_quality/test_base.py
```diff
@@ -23,14 +23,14 @@
 
 
 class TestEvolQuality:
-    def test_with_errors(self, dummy_llm: LLM) -> None:
+    def test_with_errors(self, caplog, dummy_llm: LLM) -> None:
```
alvarobartt, line 26:
```suggestion
    def test_with_errors(self, caplog: pytest.LogCaptureFixture, dummy_llm: LLM) -> None:
```

tests/unit/steps/tasks/test_base.py
```diff
@@ -66,9 +66,9 @@ def test_within_pipeline_context(self) -> None:
             assert task.llm is llm
         assert task.pipeline == pipeline
 
-    def test_with_errors(self) -> None:
-        with pytest.raises(ValueError, match="Step 'task' hasn't received a pipeline"):
-            DummyTask(name="task", llm=DummyLLM())
+    def test_with_errors(self, caplog) -> None:
```
alvarobartt, line 69:
```suggestion
    def test_with_errors(self, caplog: pytest.LogCaptureFixture) -> None:
```

tests/unit/steps/test_base.py
```diff
@@ -88,9 +88,11 @@ def test_create_step_within_pipeline_context(self) -> None:
 
         assert step.pipeline == pipeline
 
-    def test_creating_step_without_pipeline(self) -> None:
-        with pytest.raises(ValueError, match="Step 'dummy' hasn't received a pipeline"):
-            DummyStep(name="dummy")
+    def test_creating_step_without_pipeline(self, caplog) -> None:
```
alvarobartt, line 91:
```suggestion
    def test_creating_step_without_pipeline(self, caplog: pytest.LogCaptureFixture) -> None:
```

tests/unit/steps/test_base.py
```diff
@@ -284,16 +286,16 @@ def test_step_from_dict(self) -> None:
...
-            DummyStep.from_dict(
-                {
-                    **{
-                        "name": "dummy",
-                        TYPE_INFO_KEY: {
-                            "module": "tests.pipeline.step.test_base",
-                            "name": "DummyStep",
-                        },
-                    }
+    def test_step_from_dict_without_pipeline_context(self, caplog) -> None:
```
alvarobartt, line 289:
```suggestion
    def test_step_from_dict_without_pipeline_context(self, caplog: pytest.LogCaptureFixture) -> None:
```

alvarobartt (commented):
LGTM! Only some minor comments on some type-hints and would be nice to merge `develop` (i.e. `git pull origin develop`) into your branch before merging! 👍🏻

src/distilabel/steps/base.py
```diff
@@ -113,13 +113,15 @@ def model_post_init(self, __context: Any) -> None:
...
                 f"Step '{self.name}' hasn't received a pipeline, and it hasn't been"
                 " created within a `Pipeline` context. Please, use"
                 " `with Pipeline() as pipeline:` and create the step within the context."
             )
-
-        self.pipeline._add_step(self)
+        if self.pipeline is not None:
+            # If this is not set, errors will raise, but simplifies testing steps without a Pipeline.
+            self.pipeline._add_step(self)
```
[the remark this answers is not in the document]
plaguss: Could you take a second look at this suggestion? This was a source of conflict with develop, and once solved I'm not sure it it makes sense other than the comment update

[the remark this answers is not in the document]
alvarobartt: Oh sure, I'll double check the current file then and submit a new suggestion if applicable 👍🏻

alvarobartt (approved):

PR #568, opened 2024-04-24
Dump logs within a file in .cache/distilabel/pipelines dir

### LINKED ISSUE #567: Dump logs within a file in `.cache/distilabel/pipelines` dir
## Description

We could write the logs to a file in the cache dir to push it to the dataset in the hub like we do with the `pipeline.yaml`.

plaguss:
## Description

Dump the logs to a file under the `.cache/distilabel/pipelines` to allow pushing the file to the dataset in the Hugging Face hub.

A new file will be generated under the cache folder of a pipeline `pipeline.log` that will contain the same logs that are passed to the console and with the same format.

An example log file can be seen at this [dummy dataset](https://huggingface.co/datasets/distilabel-internal-testing/test-file-upload/blob/main/pipeline.log).

Closes #567

### DIFF
--- a/src/distilabel/pipeline/local.py
+++ b/src/distilabel/pipeline/local.py
@@ -107,6 +107,7 @@ class Pipeline(BasePipeline):
             return create_distiset(
                 self._cache_location["data"],
                 pipeline_path=self._cache_location["pipeline"],
+                log_filename_path=self._cache_location["log_file"],
             )
 
         buffer_data_path = self._cache_location["data"]
@@ -151,7 +152,9 @@ class Pipeline(BasePipeline):
 
         write_buffer.close()
         distiset = create_distiset(
-            self._cache_location["data"], pipeline_path=self._cache_location["pipeline"]
+            self._cache_location["data"],
+            pipeline_path=self._cache_location["pipeline"],
+            log_filename_path=self._cache_location["log_file"],
         )
         stop_logging()
         return distiset
--- a/src/distilabel/pipeline/local.py
+++ b/src/distilabel/pipeline/local.py
@@ -55,9 +55,9 @@ _STEPS_FINISHED_LOCK = threading.Lock()
 _SUBPROCESS_EXCEPTION: Union[Exception, None] = None
 
 
-def _init_worker(queue: "Queue[Any]", log_filename: str = "pipeline.log") -> None:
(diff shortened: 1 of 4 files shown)


src/distilabel/utils/logging.py
```diff
@@ -57,6 +59,11 @@ def setup_logging(log_queue: "Queue[Any]") -> None:
         handler.setFormatter(formatter)
         queue_listener = QueueListener(log_queue, handler, respect_handler_level=True)
         queue_listener.start()
+        # Given that we always append to the log file, remove it from the main process,
+        # this way we ensure every pipeline starts with a fresh log file.
+        log_filename = Path(filename)
+        if log_filename.exists():
+            log_filename.unlink()
```
gabrielmbmb, line 66: I think is better to have the default append logic, right? It will also cover the case you execute the pipeline, you stop it, then you execute it again and continue from the point it was stopped, and you have the log messages from the two executions that produced the final dataset.

src/distilabel/pipeline/local.py
```diff
@@ -55,9 +55,9 @@
 _SUBPROCESS_EXCEPTION: Union[Exception, None] = None
 
 
-def _init_worker(queue: "Queue[Any]") -> None:
+def _init_worker(queue: "Queue[Any]", log_filename: str = "pipeline.log") -> None:
     signal.signal(signal.SIGINT, signal.SIG_IGN)
-    setup_logging(queue)
+    setup_logging(queue, filename=log_filename)
```
gabrielmbmb, line 60: I would just leave `setup_logging(queue)` because like this it seems that the child processes needs also to setup the log file, when they don't have to anything with the log file.

gabrielmbmb (changes requested):
Hi, looks good to me, just a few small comments. 

It would be nice to include also the time in the log file, so you can check the log file and see how much it took.

src/distilabel/pipeline/local.py
plaguss: totally!

plaguss:
All changes done, a new log file can be seen [here](https://huggingface.co/datasets/distilabel-internal-testing/test-file-upload/blob/main/pipeline.log).

gabrielmbmb (approved):

Issue #576, opened 2024-04-24
[BUG] PreferenceToArgilla fails with custom input_mappings


dvsrepo:
**Describe the bug**
Maybe [here](https://github.com/argilla-io/distilabel/blob/712d32c70a6ba358d0a33984bcf167afcd46383a/src/distilabel/steps/argilla/preference.py#L233C50-L233C70) we should use `input[self._generations]`?

**To Reproduce**
Code to reproduce
```python
    push_to_argilla = PreferenceToArgilla(
        name="push_to_argilla",
        api_url=userdata.get('RG_DEMO_API_URL'),
        api_key=userdata.get('RG_DEMO_API_KEY'),  # type: ignore
        dataset_name="ultrallama3",
        dataset_workspace="admin",
        input_columns={"generations": "generation"},
        num_generations=2,
    )
```

**Expected behaviour**

**Screenshots**

**Desktop (please complete the following information):**
 - Package version:
 - Python version:

**Additional context**

dvsrepo:
> It should be fine as is, what's the issue exactly? Could you share the error output here? Thanks in advance 👍🏻

My column is called `generation`:

```python
push_to_argilla' failed. Sending empty batch...                                      
                    WARNING  ['distilabel.step.push_to_argilla'] Subprocess traceback:                 [local.py](file:///usr/local/lib/python3.10/dist-packages/distilabel/pipeline/local.py):[813](file:///usr/local/lib/python3.10/dist-packages/distilabel/pipeline/local.py#813)
                                                                                                                   
                             Traceback (most recent call last):                                                    
                               File                                                                                
                             "/usr/local/lib/python3.10/dist-packages/distilabel/pipeline/local.py",               
                             line 800, in _non_generator_process_loop                                              
                                 result = next(self.step.process_applying_mappings(*batch.data))                   
                               File                                                                                
                             "/usr/local/lib/python3.10/dist-packages/distilabel/steps/base.py", line              
                             391, in process_applying_mappings                                                     
                                 for output_rows in generator:                                                     
                               File                                                                                
                             "/usr/local/lib/python3.10/dist-packages/distilabel/steps/argilla/prefere             
                             nce.py", line 233, in process                                                         
                                 for idx, generation in enumerate(input["generations"])  # type:                   
                             ignore                                                                                
                             KeyError: 'generations'
```

alvarobartt:
[...]

PR #577, opened 2024-04-24
Set `extra="forbid"` in `_Step.model_config`

alvarobartt:
## Description

This PR sets the `extra="forbid"` value within the `model_config` of `_Step` so as to disallow the usage of extra attributes / fields not defined within the original `pydantic.BaseModel`.

The original idea of adding this PR was based on the issue #576, since `input_columns` was being used instead of `input_mappings`, but no exception was being raised, so the issue was not easy to identify without the `ValidationError`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
           else
             pip install -e .[dev,tests,cohere,hf-transformers,hf-inference-endpoints,vertexai,ollama,openai,together,argilla,llama-cpp,anthropic,litellm]
           fi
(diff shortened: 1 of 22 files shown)


gabrielmbmb (approved):
LGTM! maybe we can add the extra forbid to the `LLM` base class too

dvsrepo:
> LGTM! maybe we can add the extra forbid to the `LLM` base class too

Maybe we can add the extra forbid to me too 🤣

gabrielmbmb:
@dvsrepo 😂

PR #585, opened 2024-04-25
Add `stop` and `stop_sequences` in `LLM.generate` subclasses

alvarobartt:
## Description

This PR adds both `stop` to `OpenAILLM` and all their subclasses in consequence, `stop_sequences` to `InferenceEndpointsLLM`, and sets the `stop_token_ids` if not present in `vllm.SamplingParams`, all to match the existing EOS token ID if available.

If not set, both vLLM and TGI tend to eventually generate until `max_new_tokens`, which is not intended in most of the cases, so setting the EOS token to let the generation know when to stop is highly recommended.

### DIFF
--- a/src/distilabel/llms/huggingface/inference_endpoints.py
+++ b/src/distilabel/llms/huggingface/inference_endpoints.py
@@ -327,10 +327,6 @@ class InferenceEndpointsLLM(AsyncLLM):
         Returns:
             A list of lists of strings containing the generated responses for each input.
         """
-
-        if stop_sequences is None and self._tokenizer is not None:
-            stop_sequences = [self._tokenizer.eos_token]
-
         if stop_sequences is not None:
             if isinstance(stop_sequences, str):
                 stop_sequences = [stop_sequences]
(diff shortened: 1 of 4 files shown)


src/distilabel/llms/huggingface/inference_endpoints.py
```diff
@@ -316,11 +320,28 @@ async def agenerate(  # type: ignore
...
+            stop_sequences: either a single string or a list of strings containing the sequences
+                to stop the generation at. Defaults to `None`, but will be set to the
+                `tokenizer.eos_token` if available.
 
         Returns:
             A list of lists of strings containing the generated responses for each input.
         """
 
+        if stop_sequences is None and self._tokenizer is not None:
+            stop_sequences = [self._tokenizer.eos_token]
```
gabrielmbmb, line 332: Do we need this? In theory, default `stop_sequences` should be loaded automatically, right?

src/distilabel/llms/vllm.py
```diff
@@ -149,6 +149,10 @@ def generate(  # type: ignore
         """
         if extra_sampling_params is None:
             extra_sampling_params = {}
+
+        if "stop_token_ids" not in extra_sampling_params:
+            extra_sampling_params["stop_token_ids"] = [self._tokenizer.eos_token_id]
```
gabrielmbmb, line 154: same

gabrielmbmb (approved):

PR #586, opened 2024-04-25
Add `distilabel_meta` column to the datasets to include general data

### LINKED ISSUE #582: [FEATURE] Enable debugging cols similar to the previous raw_generation and labelling_prompt

I'm testing a preference pipeline with Llama3 and the parsed outputs are weird (long lists of ratings and rationales when my number of generations is two). I'd like to have a way to automatically enable some columns in my resulting distiset to inspect:

1. The prompt sent by the UltraFeedback task
2. The raw responses parsed with the UF task

This is of course applicable to other Steps but it's more crucial when using tasks with complex prompt templates and output parsing.

Something like enable debug columns when I setup the Task

Please suggest alternatives I can use in the meantime

**Additional context**

plaguss:
## Description

This PR adds a new field `distilabel_meta` to store general outputs related to distilabel. In the case of the `Tasks` it will include the `raw_output` from the step (called `f"raw_output_{self.name}"`) to include the raw output in case of failure of `self.format_output`:

```sh
# Sample dataframe of the final distiset

   model_name           generation                                distilabel_meta
0        test                 None               {'raw_output_dummy': 'output'}
1        test  llm response 0.0973     {'raw_output_dummy': 'output'}
2        test   llm response 0.341      {'raw_output_dummy': 'output'}
3        test  llm response 0.1632     {'raw_output_dummy': 'output'}
4        test                 None               {'raw_output_dummy': 'output'}
5        test  llm response 0.0733     {'raw_output_dummy': 'output'}
6        test  llm response 0.4824    {'raw_output_dummy': 'output'}
7        test                 None               {'raw_output_dummy': 'output'}
8        test  llm response 0.4292    {'raw_output_dummy': 'output'}
9        test  llm response 0.3858    {'raw_output_dummy': 'output'}
```

Closes #582


alvarobartt (commented):
LGTM! But I would rename the column to `distilabel_metadata` instead, and maybe add a flag to control whether we want to have that column in the final dataset or not? i.e. in some processes as `TextGeneration` that only adds more duplicated data while leading to a x2 bigger dataset so IMO this should be disabled by default and enabled via a flag. Optionally, we can store those in disk in the cache under a separate file in any case, but I wouldn't add that by default to the `distiset` unless e.g. `enable_metadata=True`, WDYT? @plaguss

plaguss:
I've added two different options to include the metadata.
One at the `Task` level:

```python
TextGeneration(..., add_raw_output=True | False)
```

So we can determine if we want that specific field to be present, assuming we will have more fields in the future inside `distilabel_metadata`.

Another, at the `Pipeline` level:

```python
with Pipeline(..., enable_metadata=True | False):
    ...
```

This way we can decide to remove all the column altogether.

PR #590, opened 2024-04-26
Fix empty batches causing missaligment when branching

gabrielmbmb:
## Description

This PR fixes an issue that caused batch miss-alignment in pipeline with steps like:  A -> [B, C] -> D. If for some reason, C or D failed, then we were sending an empty list causing creating miss-aligned batches for step D. To fix it, instead of sending an empty list, now we send the data of the input batch + the `outputs` of the step as new keys with `None` as value.

### DIFF
--- a/tests/integration/test_branching_missaligmnent.py
+++ b/tests/integration/test_branching_missaligmnent.py
@@ -1,0 +1,55 @@
+# Copyright 2023-present, Argilla, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


tests/integration/test_branching_missaligmnent.py
```diff
@@ -0,0 +1,55 @@
...
+
+        fail = FailAlways()
+        succeed = SucceedAlways()
+        combine = CombineColumns(columns=["response"])
+
+        load_data.connect(fail)
+        load_data.connect(succeed)
+
+        fail.connect(combine)
+        succeed.connect(combine)
```
plaguss, line 48:
```suggestion
        load_data >> [fail, succeed] >> combine
```

PR #603, opened 2024-05-03
Fix `num_examples` to be optional in `LoadHubDataset`

### LINKED ISSUE #602: [BUG] `num_examples` is a mandatory kwarg in `LoadHubDataset`
**Describe the bug**

`num_examples` in `LoadHubDataset` is a new and mandatory parameter, but it should be optional and default to `None`.

**To Reproduce**

Really any `Pipeline` with a `LoadHubDataset` step without the value `num_examples` set neither in the `__init__` nor as a runtime parameter in the `parameters` arg in `Pipeline.run`.

**Expected behaviour**

`num_examples` to be an optional value, since in most of the cases we want to run something over the whole dataset, and having to specify `num_examples=None` doesn't feel rigth.

**Screenshots**

image

**Desktop:**
 - Package version: `develop`
 - Python version: 3.11

plaguss:
## Description

Fix a bug that makes `num_examples` mandatory in `LoadHubDataset`.

Closes #602.

### DIFF
--- a/src/distilabel/steps/generators/huggingface.py
+++ b/src/distilabel/steps/generators/huggingface.py
@@ -109,7 +109,7 @@ class LoadHubDataset(GeneratorStep):
         default=False,
         description="Whether to load the dataset in streaming mode or not. Defaults to False.",
     )
-    num_examples: RuntimeParameter[int] = Field(
+    num_examples: Optional[RuntimeParameter[int]] = Field(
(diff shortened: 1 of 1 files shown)


alvarobartt (approved):

alvarobartt:
Mate is you day off, disconnect 😠 @plaguss

PR #604, opened 2024-05-03
Add the option of passing the multiprocessing context via env var

plaguss:
## Description

This PR adds the option of passing the multiprocessing backend via environment variable, overwriting the default:

`DISTILABEL_MP_CONTEXT` can be one of the [`multiprocessing`](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) available ones.

There is no issue related, but apparently `llama-cpp-python` has issues with multiprocessing, `spawn` seems to work in some computers and `forkserver` doesn't work always.


gabrielmbmb:
If `spawn` works for all the OS (I think so) and it also works with all the LLMs, then we can set it as the default one.

plaguss:
Following our internal discussion I checked that one can just set the start method using multiprocessing, so there's no need for the env var.

```python
if __name__ == "__main__":
    import multiprocessing as mp
    mp.set_start_method("spawn")

    with Pipeline("my-pipe") as pipeline:
        ...
```

 Also the error from `llama-cpp` is no longer happening, so we can close this one (letting the defaults as we have them, because there aren't more errors to control).

gabrielmbmb:
@plaguss we need to call the `get_context` function without any argument so it uses the one set in `mp.set_start_method`

https://github.com/argilla-io/distilabel/blob/6af680d19356d82e39e11934ff72191ee6cae068/src/distilabel/pipeline/local.py#L135

src/distilabel/pipeline/local.py
```diff
@@ -132,7 +121,7 @@ def run(
         write_buffer = _WriteBuffer(buffer_data_path, self.dag.leaf_steps)
 
         num_processes = len(self.dag)
-        ctx = mp.get_context(_MULTIPROCESSING_CONTEXT)  # type: ignore
+        ctx = mp.get_context()  # type: ignore
```
alvarobartt, line 124: Won't this fail if no `start_method` is set?

gabrielmbmb: no, it uses the default one

PR #620, opened 2024-05-08
Update `UltraFeedback` paper replication to use `routing_batch_function`

gabrielmbmb:
## Description

This PR updates `UltraFeedback` paper replication to use `routing_batch_function` to replicate the LLM pooling mechanism described in the paper.


docs/sections/papers/ultrafeedback.md
```diff
@@ -46,43 +47,66 @@ from distilabel.steps import (
 )
 from distilabel.steps.tasks import TextGeneration, UltraFeedback
 
+sample_three_llms = sample_n_steps(3)
```
alvarobartt, line 50: Maybe use the kwarg here?

docs/sections/papers/ultrafeedback.md
```diff
@@ -46,43 +47,66 @@ from distilabel.steps import (
...
+
 
 with Pipeline(name="ultrafeedback-pipeline") as pipeline:
     load_hub_dataset = LoadHubDataset(
         name="load_dataset",
         output_mappings={"prompt": "instruction"},
+        batch_size=2,
```
alvarobartt, line 57: Does this really matter if we're already setting `input_batch_size` in the follow up steps?

docs/sections/papers/ultrafeedback.md
```diff
@@ -46,43 +47,66 @@ from distilabel.steps import (
...
         output_mappings={"model_name": "generation_model"},
     )
-    load_hub_dataset.connect(text_generation_with_notus)
-    load_hub_dataset.connect(text_generation_with_zephyr)
 
     combine_columns = CombineColumns(
         name="combine_columns",
         columns=["generation", "generation_model"],
         output_columns=["generations", "generation_models"],
+        input_batch_size=2
```
alvarobartt, line 101: This shouldn't be required right?

docs/sections/papers/ultrafeedback.md
alvarobartt, line 50: Also, can we include that within the `Pipeline` context manager? Or does it need to be defined outside? If there's any reason for that maybe adding a code comment here would be nice!

docs/sections/papers/ultrafeedback.md
alvarobartt: Assuming this is only for the `sample_n_steps` function, shouldn't we define the `batch_size` within that function instead of on the previous one?

docs/sections/papers/ultrafeedback.md
gabrielmbmb: It is. By default the `input_batch_size` is 50, but as this is a convergence step (it will receive batches of steps that received routed batches) it has to be lower or equal than the `input_batch_size` of the previous steps.

docs/sections/papers/ultrafeedback.md
gabrielmbmb: It doesn't matter where you use the function, as a `RoutingBatchFunction` doesn't have/need a `pipeline` argument.

gabrielmbmb: It's similar to when you create a new `Step`

docs/sections/papers/ultrafeedback.md
gabrielmbmb: Yes, it matters. The routing batch function, as the name indicates, routes batches. If we have the default `batch_size` of 50, then we will have much less batches, and therefore much less variety in the LLMs that were chosen.

docs/sections/papers/ultrafeedback.md
gabrielmbmb: Okay xD

gabrielmbmb, line 50:
```suggestion
sample_three_llms = sample_n_steps(n=3)
```

alvarobartt: Oh fair, but in this case it "doesn't need to be" or "cannot be" inside the context manager?

docs/sections/papers/ultrafeedback.md
alvarobartt: Then does that mean that if in one step we define `input_batch_size=1` we are forcing the rest of the steps below in the pipeline to use the same batch size?

docs/sections/papers/ultrafeedback.md
alvarobartt: Yes, what I meant is whether isn't it more intuitive to define the output batch size of the routing in the `sample_n_steps` function instead, as that function will receive a batch of inputs of N size, but can then route those in mini batches, say of size N // 5, right? I asked more in terms of how intuitive or counterintuitive can be right now.

alvarobartt (approved):

PR #630, opened 2024-05-13
Force pipeline stop after 2 SIGINT signals caught

plaguss:
## Description

Force pipeline to stop after 2 `ctrl+c` presses:

image

Closes #406

### DIFF
--- a/src/distilabel/pipeline/local.py
+++ b/src/distilabel/pipeline/local.py
@@ -663,16 +663,16 @@ class Pipeline(BasePipeline):
             if _STOP_CALLED:
                 global _STOP_CALLS
                 _STOP_CALLS += 1
+                # if _STOP_CALLS == 1:
+                #     self._logger.warning(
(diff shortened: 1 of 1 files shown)


src/distilabel/pipeline/local.py
```diff
@@ -658,10 +659,28 @@ def _stop(self) -> None:
...
+                else:
+                    self._logger.warning("🛑 Forcing pipeline interruption.")
+                    import gc
+                    import sys
+
+                    if manager:
+                        manager.shutdown()
+                    gc.collect()
+
+                    sys.exit(1)
```
gabrielmbmb, line 682: I think we can force the stop at second CTRL+C and updating the first log message to include something like: "Press again to force pipeline stop" should be enough.

gabrielmbmb (changes requested):

src/distilabel/pipeline/local.py
plaguss: Done

PR #640, opened 2024-05-15
Export components info & components gallery in docs

gabrielmbmb:
## Description

This PR adds an auxiliary function `export_components_info` that can be used to get a dictionary containing all the relevant information about the components (Steps, Tasks and LLMs) that can be used with `distilabel`. This function can be used in 3rd party applications to generate apps on top of `distilabel` (UI for example 👀).

This function is used by the `distilabel/components-gallery` mkdocs plugin also included in this PR, which automatically creates a component gallery in distilabel docs.


alvarobartt:
Maybe point to `update-docs` instead?

plaguss (approved):

plaguss (approved):

PR #649, opened 2024-05-18
Fix routing batch function deadlocks and unordered batches

gabrielmbmb:
## Description

This PR fixes several issues that caused deadlocks and having a final dataset with unordered batches caused when a routing batch function was used in a pipeline. Several tests have been added to check that everything is working correctly.

### DIFF
--- a/src/distilabel/pipeline/base.py
+++ b/src/distilabel/pipeline/base.py
@@ -545,6 +545,16 @@ class _BatchManagerStep(_Serializable):
         convergence_step: A flag to indicate if the step is a convergence step. An
             `Step` is a convergence step if all its predecessors are receiving routed
             batches. Defaults to `False`.
+        convergence_step_batches_consumed: A dictionary in which the key is the `seq_no`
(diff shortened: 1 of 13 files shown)


src/distilabel/pipeline/base.py
```diff
@@ -428,12 +428,6 @@ def _load_from_cache(self) -> None:
         """
         cache_loc = self._cache_location
         if cache_loc["pipeline"].exists():
-            # Refresh the DAG to avoid errors when it's created within a context manager
-            # (it will check the steps aren't already defined for the DAG).
-            self.dag = DAG()
-            new_class = self.from_yaml(cache_loc["pipeline"])
-            # Update the internal dag and batch_manager
-            self.dag.G = new_class.dag.G
```
plaguss, line 436: we don't need to update the internal `DAG` anymore?

plaguss (commented):
LGTM! Just a comment, but if the tests are passing and you are sure of it, we are good to go

PR #655, opened 2024-05-21
Fix error on `Pipeline.dry_run` without `parameters`

plaguss:
## Description

This PR fixes a bug (thanks @sdiazlor) that prevented calling `dry_run` method without `parameters` argument informed.


alvarobartt (approved):
Thanks for the bug report @sdiazlor and for the quick fix @plaguss 🤗

sdiazlor (approved):
Thanks :)

PR #658, opened 2024-05-21
Fix crash when using vLLM without structured generation

cg123:
When using the vLLM backend, currently structured generation must be set or the task will crash with the following error:

```
 "/usr/local/lib/python3.10/dist-packages/vllm/model_executor/layers/logits_processor.py", line 106, in _apply_logits_processors
    logits_row = logits_processor(token_ids, logits_row)
TypeError: 'NoneType' object is not callable
```

This is because `[self._logits_processor]` is passed to vLLM as the list of logit processors, even when `self._logits_processor` is `None`. vLLM isn't picky and happily calls that `None`.

Just a one-line tweak to make things work again.

### DIFF
--- a/src/distilabel/__init__.py
+++ b/src/distilabel/__init__.py
@@ -14,6 +14,6 @@
 
 from rich import traceback as rich_traceback
 
-__version__ = "1.1.0"
+__version__ = "1.2.0"
 
 rich_traceback.install(show_locals=True)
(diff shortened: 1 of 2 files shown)


alvarobartt:
Thanks for the quick fix @cg123! We'll release `distilabel` v1.1.1 tomorrow with this, thanks!

alvarobartt (approved):

Issue #538, opened 2024-04-16
[FEATURE] OpenAI Batch API Support


jphme:
See [here](https://platform.openai.com/docs/api-reference/batch/create) for the API and [here](https://twitter.com/jeffintime/status/1779924149755924707?t=Tmo3Aoo62N5zA5PWLNiqUw) for the (Twitter) announcement.

50% discount would be huge as most large jobs running on distilabel are not time-critical for us.

However there might be some architectural decisions (should a pipeline just loop infinitely and wait for results?) around implementing this. 

If you agree that this should be in distilabel and we fletched out how this fits into the API, our team could probably do the implementation.

gabrielmbmb:
Hi @jphme!

The current architecture of `distilabel` assumes that `LLM.generate` is a blocking method that returns the generations right away.

Having that said, I think that we could have a `GeneratorStep` that uploads a file with the requests and send the request to create the batch job for that file. If the output file gets updated for every request finished, then this step could be pooling the output file and yield batches with the response for the requests of the batch that have finished.

If the output file doesn't get updated when a request of the batch job have finished, then we could have a `LoadOpenAIBatchResults` generator step that checks if a batch job have finished and if it has, then it reads the output file and yields batches of the specified size for the rest of the pipeline.

PR #664, opened 2024-05-23
Add OpenAI API Batch Processing Step

bjoernpl:
This PR adds a general step that enables the use of the OpenAI Batch API as discussed in #538. The Step follows roughly the same API as a Task but with some small changes to how the llm and generation kwargs etc. are set.

I'm open to suggestions on how this can be improved or made clearer :)

## Overview
The AnswerBatch class facilitates asynchronous processing of large amounts of requests using the OpenAI Batch API. This functionality is designed to handle batch jobs that may take up to 24 hours to complete, leveraging the OpenAI API for efficient large-scale processing with 50% lower costs.

## Key Features
- Asynchronous batch processing using the OpenAI API.
- Configurable model and generation parameters.
- Polling mechanism to check batch job status until completion.
- Abstract methods for input and output formatting, allowing customization for various tasks.

### Basic Implementation

1. **Define a Custom Batch Class**

   Create a class inheriting from `AnswerBatch` and implement the `format_input` and `format_output` methods to suit your task requirements. Similar to creating a Task for processing with an LLM.

   ```python
   class AnswerWithBatchAPI(AnswerBatch):
       @override
       def format_input(self, input: Dict[str, Any]) -> List[ChatItem]:
           return [{"role": "user", "content": input["instruction"]}]
   
       @override
       def format_output(self, output: str | None, input: Dict[str, Any]) -> Dict[str, Any]:
           return {
               "response": output,
           }
   
       @property
           return ["instruction"]
   
       @property
           return ["response"]
   ```

2. **Setup and Run Pipeline**

   Define a pipeline that includes loading data, renaming columns, and generating responses using the batch processing class.

   ```python
   from distilabel.steps.generators.data import LoadDataFromDicts

   DATA = [
       {"instruction": "Tell me a joke"},
       {"instruction": "Write a short haiku"},
       # Add more prompts as needed
   ]

   def run_pipeline():
       with Pipeline(name="unit-test-pipeline") as pipeline:
           load_dataset = LoadDataFromDicts(name="load_dataset", data=DATA)
           generate_response = AnswerWithBatchAPI(
               name="generate_response", model_id="gpt-4o", batch_description="Testing OpenAI batch processing"
           )

           load_dataset.connect(generate_response)

       return pipeline.run(
           parameters={}
       )

   def test_pipeline_cached():
       ds = run_pipeline()
       assert len(ds["default"]["train"]) == 80

   if __name__ == "__main__":
       test_pipeline_cached()
   ```

### Configuration Options

- `model_id`: The ID of the OpenAI model to use (e.g., "gpt-4o").
- `generation_kwargs`: Dictionary of generation parameters (e.g., `max_tokens`, `temperature`).
- `openai_client_kwargs`: Additional arguments for initializing the OpenAI client.
- `batch_description`: Description for the batch job.
- `continue_batch`: Boolean to indicate if the batch should continue from a previous job.
- `poll_every_n_seconds`: Interval for polling the batch job status.
- `input_batch_size`: Size of the input batch to process.

### Methods

- `format_input`: Abstract method to format inputs for the batch job.
- `format_output`: Abstract method to format outputs from the batch job.
- `process`: Main method to handle the batch processing flow, including creating batches, polling status, and processing results.

### DIFF
--- a/src/distilabel/steps/globals/answer_batch.py
+++ b/src/distilabel/steps/globals/answer_batch.py
@@ -1,0 +1,221 @@
+import json
+import time
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+
+from typing_extensions import override
+
+from distilabel.steps import GlobalStep, StepOutput
+from distilabel.steps.base import StepInput
+from distilabel.steps.tasks.typing import ChatType
+
+if TYPE_CHECKING:
+    from distilabel.llms.typing import GenerateOutput
+    from distilabel.steps.tasks.typing import ChatType
+    from distilabel.steps.typing import StepOutput
+
+from openai import OpenAI
+from openai.types.batch import Batch
+
+
+class AnswerBatch(GlobalStep, ABC):
+    """Use the OpenAI Batch API to process a large amount of requests asynchronously.
+    Requests may take up to 24 hours to complete. See https://platform.openai.com/docs/guides/batch/getting-started for more info on the batch API.
+    This Step behaves exactly like a task would except you set the model_id and generation_kwargs as separate attributes instead of in the LLM definition.
+
+    It works by sending a batch request to the API and then polling for maximum 24 hours until the batch job is completed. The results are then processed and returned.
+    """
+
+    model_id: str = "gpt-4o"
+    generation_kwargs: Dict[str, Any] = {"max_tokens": 4096, "temperature": 0.7}
+    openai_client_kwargs: Dict[str, Any] = {}
+    batch_description: Optional[str] = "Processing Distilabel Step"
+    continue_batch: bool = False
+    poll_every_n_seconds: int = 10
+    batch_id: Optional[str] = None
+    batch_metadata: Dict[str, Any] = {}
+    input_batch_size: int = 10000000  # process all documents in one batch
+    _client: OpenAI = None
+
+    @abstractmethod
+    def format_input(self, input: Dict[str, Any]) -> "ChatType":
+        """Abstract method to format the inputs of the task. It needs to receive an input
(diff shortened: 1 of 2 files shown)


bjoernpl:
Still in Draft because there are some things I couldn't fully test yet since I always need to wait for the results...

An open issue that might occur is that the polling breaks when internet connection is lost since try except is still checking for a general exception and not the specific "Batch doesn't exist" exception.
Tests also don't cover restarting with a given batch_id nor changing the generation kwargs etc. Open for contributions and comments on this side @plaguss @gabrielmbmb :)

bjoernpl:
Just seeing now that tests are failing (aside from the py3.10 formatting, will fix) because inferencing openai API is not intended in the integration tests. What would be the best way to do these tests then?

gabrielmbmb:
Hi @bjoernpl, thanks for the PR!

We were planning to tackle this with a different approach, trying to avoid polling. The idea was to introduce a feature in the `GlobalStep` that allows to execute some offline job (OpenAI API Batch request), store the information needed to check the status of the job in the cache and after that stop the pipeline. If the pipeline gets executed again, it will jump directly to the `GlobalStep`, and on this second execution it will check the status of the job instead of sending the request again. If the job has finished the pipeline execution will continue, otherwise, the pipeline will be stopped again.

This removes the polling, but it's true that the user will be in charge on making that pooling. In the future, if OpenAI allows notifications of finished jobs via webhook + the addition of the `distilabel` server (we will have an endpoint to receive this kind of notifications), this will be much more sophisticated and the pipeline execution could be resumed automatically.

Anyways, we haven't started developing this feature I describe you, so your opinion on this is very valuable.

Thanks!

gabrielmbmb:
Close in favour of #923, which will be release in `1.4.0`.

PR #670, opened 2024-05-23
Add `ArenaHard` benchmark and `ArenaHardResults` step

alvarobartt:
## Description

This PR adds the `ArenaHard` task to evaluate the completions previously generated by two LLMs using the Arena Hard templates, to provide both an alternate response to the given instruction besides the two provided, plus an additional field with the score as A>B, A>>B, A=B, B>A, and B>>A.

Additionally, in order to show the ELO result as the output of the benchmark itself, this PR also includes the `ArenaHardResults` step which is a `GlobalStep` that ideally should be run right after the `ArenaHard` step, and will calculate the ELO for the given models; ideally a reference model like GPT-4 should be used for comparison.

To explore the dataset further as well as some sample generations and evaluations check https://huggingface.co/spaces/lmsys/arena-hard-browser

## Resources

To run the `ArenaHard` benchmark using `distilabel`, we've ported the `question.jsonl` with the questions to generate completions for, from [`lm-sys/arena-hard-auto/data/arena-hard-v0.1/question.jsonl`](https://github.com/lm-sys/arena-hard-auto/blob/main/data/arena-hard-v0.1/question.jsonl) into a repository in the Hugging Face Hub at [`alvarobartt/lmsys-arena-hard-v0.1`](https://huggingface.co/datasets/alvarobartt/lmsys-arena-hard-v0.1) (still pending approval / permission from the authors).

## References

- https://github.com/lm-sys/arena-hard-auto/tree/main
- https://lmsys.org/blog/2024-04-19-arena-hard/


src/distilabel/steps/tasks/benchmarks/arena_hard.py
```diff
@@ -0,0 +1,59 @@
...
+        if output is None:
+            return {"evaluation": None, "result": None}
+        pattern = re.compile(r"\[\[([AB<>=]+)\]\]")
+        match = pattern.search(output)
+        if match is None:
+            return {"evaluation": output, "result": None}
+        return {"evaluation": output, "result": match.group(1)}
+
+
+class ArenaHardResults(GlobalStep):
```
ignacioct, line 56: Is this intended to be empty or is it WIP?

alvarobartt: Yes, it's still WIP since I was double checking that still, I'll push that later today ideally! 👍🏻

alvarobartt:
P.S. I'd still like to include an example under the `examples/` directory so as to showcase how to re-use the GPT-4 existing answers to avoid generating those, as those are the reference answers atm and doesn't make much sense to generate those over and over.

src/distilabel/steps/tasks/benchmarks/arena_hard.py
```diff
@@ -0,0 +1,230 @@
...
+
+from distilabel.steps import GlobalStep, StepInput
+from distilabel.steps.tasks.base import Task
+from distilabel.steps.tasks.typing import ChatType
+from distilabel.steps.typing import StepOutput
+
+
+class ArenaHard(Task):
+    """This `Task` is based on the "From Live Data to High-Quality Benchmarks: The
+    Arena-Hard Pipeline" paper, which presents Arena Hard which is a benchmark for
```
ignacioct, line 26: `which` used two times in the same sentence

src/distilabel/steps/tasks/benchmarks/arena_hard.py
```diff
@@ -0,0 +1,230 @@
...
+    instruction-tuned LLMs that contains 500 challenging user queries. GPT-4 is used
+    as the judge to compare the model responses against a baseline model, which defaults
+    to `gpt-4-0314`.
+
+    Note:
+        Arena-Hard-Auto has the highest correlation and separability to Chatbot Arena
+        among popular open-ended LLM benchmarks.
```
ignacioct, line 33:
```suggestion
        Arena-Hard-Auto has the highest correlation and separability of Chatbot Arena
        among popular open-ended LLM benchmarks.
```

src/distilabel/steps/tasks/benchmarks/arena_hard.py
```diff
@@ -0,0 +1,230 @@
...
+        return {"evaluation": output, "score": match.group(1)}
+
+
+class ArenaHardResults(GlobalStep):
+    """This `Step` is based on the "From Live Data to High-Quality Benchmarks: The
+    Arena-Hard Pipeline" paper, which presents Arena Hard which is a benchmark for
+    instruction-tuned LLMs that contains 500 challenging user queries. This step is
+    a `GlobalStep` that should run right after the `ArenaHard` task to calculate the
+    ELO scores for the evaluated models.
+
```
ignacioct, line 113: the same which problem

src/distilabel/steps/tasks/benchmarks/arena_hard.py
```diff
@@ -0,0 +1,230 @@
...
+    Note:
+        Arena-Hard-Auto has the highest correlation and separability to Chatbot Arena
```
ignacioct, line 115: same as above, would you use `of` instead of `to`.

ignacioct (approved):
Read the paper and the code looks nice! Added a few comments on the writing, but LGTM. Looking forward for the example.

src/distilabel/steps/tasks/benchmarks/arena_hard.py
alvarobartt:
```suggestion
    Arena-Hard Pipeline" paper that presents Arena Hard, which is a benchmark for
```

src/distilabel/steps/tasks/benchmarks/arena_hard.py
alvarobartt: I believe in this case the correct form is `to` (also this text was a copy-paste with minimal edits from their official post), so not 100% sure either

src/distilabel/steps/tasks/benchmarks/arena_hard.py
alvarobartt:
```suggestion
    Arena-Hard Pipeline" paper that presents Arena Hard, which is a benchmark for
    instruction-tuned LLMs that contains 500 challenging user queries. This step is
    a `GlobalStep` that should run right after the `ArenaHard` task to calculate the
    ELO scores for the evaluated models.

```

PR #673, opened 2024-05-24
Add functionality to load/save distisets to/from disk

plaguss:
## Description

This PR includes two new methods on `Distiset` to save the content to disk (or load from it), and also to a remote storage based on the content of the `storage_options` passed, following the implementation of Hugging Face's datasets: https://huggingface.co/docs/datasets/filesystems#saving-serialized-datasets

```python
with Pipeline(...) as pipe:
    distiset = pipe.run()

distiset.save_to_disk(distiset_path)  # Or include the storage_options for the remote storage.
ds = Distiset.load_from_disk(distiset_path)
```


src/distilabel/distiset.py
```diff
@@ -193,6 +225,193 @@ def train_test_split(
...
+                card.save(new_filename)
+
+        # Write our internal files to the distiset folder by copying them to the distiset folder.
+        if save_pipeline_config and self.pipeline_path:
+            new_filename = posixpath.join(
+                distiset_config_folder, PIPELINE_CONFIG_FILENAME
+            )
+            if self.pipeline_path.exists() and (not fs.isfile(new_filename)):
+                data = yaml.safe_load(
+                    self.pipeline_path.read_text(), Loader=yaml.FullLoader
```
rasdani, line 316: I got `TypeError: safe_load() got an unexpected keyword argument 'Loader'`
and fixed it with.
```python
                data = yaml.safe_load(
                    # self.pipeline_path.read_text(), Loader=yaml.FullLoader
                    self.pipeline_path.read_text()
                )
```
I assume for `yaml.safe_load()` one does not need to specify a `Loader` anymore

plaguss: I think I mixed `safe_load` with `load` with the arguments, thanks for the catch!

plaguss:
@rasdani did you have time to try saving/loading to s3 with this branch?

rasdani:
yes, saving and loading with S3 works so far! :)

However one needs to keep in mind, to append `default/train` (or your respective config/split name) to your S3 directory path.
This is expected behaviour from `datasets.load_from_disc()` though.

src/distilabel/distiset.py
```diff
@@ -193,6 +225,193 @@ def train_test_split(
...
+                for more information. Defaults to None.
+            storage_options: Key/value pairs to be passed on to the file-system backend, if any.
+                Defaults to None.
+
+        Returns:
+            Distiset
+        """
+        distiset_path = str(distiset_path)
+        fs: fsspec.AbstractFileSystem
+        fs, _, [distiset_path] = fsspec.get_fs_token_paths(
```
rasdani, line 348: I think, I found an issue:
when loading with a `distiset_path` that has `"s3://"`as prefix. This prefix gets discarded here.
This line turns `"s3://my/example/dataset"` into `"my/example/dataset"` and subsequent loading from S3 with `Dataset.load_from_disk()` fails. (Effectively it interprets the truncated path as an local path and fails, if no such local dataset is present.)

rasdani: I'm currently looking into, if this can be solved via an option in `storage_options`

rasdani: Screenshot 2024-05-27 at 16 29 36

The paths are missing `"s3://"` and start right away with `"example/path"`

mkdocs.yml
```diff
@@ -62,6 +62,7 @@ markdown_extensions:
   - attr_list
   - md_in_html
   - admonition
+  - codehilite
```
gabrielmbmb, line 65: is this needed for the `Examples` section in the docstrings?

src/distilabel/distiset.py
```diff
@@ -13,21 +13,31 @@
...
-from distilabel.steps.tasks.base import DISTILABEL_METADATA_KEY
 from distilabel.utils.card.dataset_card import (
     DistilabelDatasetCard,
     size_categories_parser,
 )
 from distilabel.utils.files import list_files_in_dir
 
+DISTISET_CONFIG_FOLDER: str = "distiset_configs"
+PIPELINE_CONFIG_FILENAME: str = "pipeline.yaml"
+PIPELINE_LOG_FILENAME: str = "pipeline.log"
```
gabrielmbmb, line 39: if we want to be a bit more precise with the type hint we can use `typing.Final[str]`

src/distilabel/distiset.py
```diff
@@ -193,6 +225,198 @@ def train_test_split(
...
+                `max_shard_size` and `num_proc`. Defaults to None.
+            num_proc: Number of processes when downloading and generating the dataset locally.
+                Multiprocessing is disabled by default. Defaults to None.
+            storage_options: Key/value pairs to be passed on to the file-system backend, if any.
+                Defaults to None.
+            save_card: Whether to save the dataset card. Defaults to True.
+            save_pipeline_config: Whether to save the pipeline configuration file (aka the `pipeline.yaml` file).
+                Defaults to True.
+            save_pipeline_log: Whether to save the pipeline log file (aka the `pipeline.log` file).
+                Defaults to True.
```
gabrielmbmb, line 261:
```suggestion
                Defaults to `None`.
            num_shards: Number of shards to write. By default the number of shards depends on
                `max_shard_size` and `num_proc`. Defaults to `None`.
            num_proc: Number of processes when downloading and generating the dataset locally.
                Multiprocessing is disabled by default. Defaults to `None`.
            storage_options: Key/value pairs to be passed on to the file-system backend, if any.
                Defaults to `None`.
            save_card: Whether to save the dataset card. Defaults to `True`.
            save_pipeline_config: Whether to save the pipeline configuration file (aka the `pipeline.yaml` file).
                Defaults to `True`.
            save_pipeline_log: Whether to save the pipeline log file (aka the `pipeline.log` file).
                Defaults to `True`.
```

src/distilabel/distiset.py
```diff
@@ -193,6 +225,198 @@ def train_test_split(
...
+    ) -> Self:
+        """Loads a dataset that was previously saved using `Distiset.save_to_disk` from a dataset
+        directory, or from a filesystem using any implementation of `fsspec.spec.AbstractFileSystem`.
+
+        Args:
+            distiset_path: Path ("dataset/train") or remote URI ("s3://bucket/dataset/train").
+            keep_in_memory: Whether to copy the dataset in-memory, see `datasets.Dataset.load_from_disk``
+                for more information. Defaults to None.
+            storage_options: Key/value pairs to be passed on to the file-system backend, if any.
+                Defaults to None.
```
gabrielmbmb, line 342:
```suggestion
            keep_in_memory: Whether to copy the dataset in-memory, see `datasets.Dataset.load_from_disk``
                for more information. Defaults to `None`.
            storage_options: Key/value pairs to be passed on to the file-system backend, if any.
                Defaults to `None`.
```

src/distilabel/distiset.py
```diff
@@ -193,6 +225,198 @@ def train_test_split(
...
+            )
+            if Path(log_filename_path).exists():
+                distiset.log_filename_path = Path(log_filename_path)
+
+        return distiset
+
+    @property
+    def pipeline_path(self) -> Optional[Path]:
+        """Contains the path to the `pipeline.yaml` file that generated the `Pipeline`."""
+        return self._pipeline_path
```
gabrielmbmb, line 405:
```suggestion
    def pipeline_path(self) -> Union[Path, None]:
        """Returns the path to the `pipeline.yaml` file that generated the `Pipeline`."""
        return self._pipeline_path
```

src/distilabel/distiset.py
```diff
@@ -193,6 +225,198 @@ def train_test_split(
...
+
+    @pipeline_path.setter
+    def pipeline_path(self, path: PathLike) -> None:
+        self._pipeline_path = Path(path)
+
+    @property
+    def log_filename_path(self) -> Optional[Path]:
+        """Contains the path to the `pipeline.log` file that generated the `Pipeline`."""
+        return self._log_filename_path
```
gabrielmbmb, line 414:
```suggestion
    def log_filename_path(self) -> Union[Path, None]:
        """Returns the path to the `pipeline.log` file that generated the `Pipeline`."""
        return self._log_filename_path
```

gabrielmbmb (approved):
LGTM!

mkdocs.yml
plaguss: That's what the docs from material say, but I've checked and we don't really need it apparently, so I'm removing it.

src/distilabel/distiset.py
plaguss: It should be fixed now @rasdani, but if you have time let me know if it works for you 😄

rasdani: @plaguss 
yes, I can load the Distiset with `"/default"` at the end of my path :)
I think this matches `datasets` behaviour for `DatasetDict`s.

rasdani: after saving once more with `Distitset.save_to_disk()` to S3, I no don't need `"/default"` at the end when loading.
Not 100% sure how `datasets` handles this, but this works for us now 👍

alvarobartt: Thanks for the confirmation @rasdani, indeed Hugging Face handles that automatically whenever the config is `default` i.e. only one config within the dataset; and when loading the dataset and multiple configs are available all the configs are downloaded and loaded. Alternatively one can also provide the `config_name` to load one config or the other. But long-story-short, the behaviour now is as intended, the `/default` or whatever config name there is, should not be provided. 🤗

PR #674, opened 2024-05-27
Add `codspeed` benchmarks

gabrielmbmb:
## Description

This PR setups [CodSpeed](https://codspeed.io/dashboard) to run and profile benchmarks. Kudos to @plaguss for suggesting having benchmarks.


plaguss (approved):

gabrielmbmb:
We can merge this PR after merging #615. `codspeed` requires `python>=3.12` for improved tracing (because in `3.12` the compatibility with the Linux perf profiler was added, and I think that's what `codspeed` uses behind the scenes)

PR #675, opened 2024-05-28
Fix circular import due to DISTILABEL_METADATA_KEY

plaguss:
## Description

This PR fixes a bug due to a circular import in `distiset.py`, to check:
```python
from distilabel.distiset import create_distiset
```

### DIFF
--- a/src/distilabel/distiset.py
+++ b/src/distilabel/distiset.py
@@ -21,7 +21,6 @@ from datasets import load_dataset
 from huggingface_hub import DatasetCardData, HfApi
 from pyarrow.lib import ArrowInvalid
 
-from distilabel.steps.tasks.base import DISTILABEL_METADATA_KEY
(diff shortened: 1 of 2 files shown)


src/distilabel/distiset.py
```diff
@@ -225,6 +224,7 @@ def create_distiset(  # noqa: C901
         correspond to different configurations of the dataset.
     """
     logger = logging.getLogger("distilabel.distiset")
+    from distilabel.steps.constants import DISTILABEL_METADATA_KEY
```
alvarobartt, line 227: Could we make this the first line within the method?

```suggestion
    from distilabel.steps.constants import DISTILABEL_METADATA_KEY
    
    logger = logging.getLogger("distilabel.distiset")
```

alvarobartt (approved):

PR #678, opened 2024-05-28
Pass data of batches using file system

### LINKED ISSUE #488: Improve passing large `_Batch` to steps
The `mp.Queue` that we're using to pass the data between the steps is very slow when it's used to send a lot of data (for example when accumulating `embeddings` from `GenerateEmbeddings` step for the `DeitaFiltering` step).

We need to develop a better way (probably using an storage and dumping the content to a file so it can be read back by the other step) to pass large batches, or using a different `Queue` implementation like this one: https://github.com/alex-petrenko/faster-fifo

gabrielmbmb:
## Description

This PR adds a new feature that allows to pass the `Batch`es between the main process and the child process using the file system (using `fsspec` abstraction). This improves the time it takes to read (from queue) a big batch, so this is specially beneficial for `GlobalStep`s receiving huge batches.

By default, the file system will be used to pass the data of the batches to `GlobalStep`s, but it can be activated for all the steps with the flag `use_fs_to_pass_data` in the `run` method. This will be specially useful for distributed pipelines in which a cloud storage such as GCS, S3, etc could be used for passing batches of bigger size between the steps. The file system to be used can be configured via the `storage_parameters` that has been added to the `BasePipeline.run` method.

Finally, some logic from `Pipeline.__init__` has been moved to `BasePipeline.__init__` so everything is more tidy and less redundant having in mind future pipelines implementations.

Closes #488


plaguss (commented):
Nice PR refactor!! just a small update related to `protocol` key in `storage_parameters`, but feel free to drop it

PR #720, opened 2024-06-11
Add `requirements` list for `Pipeline`

plaguss:
## Description

This PR adds a new attribute to `requirements` to `BasePipeline` to keep track of the dependencies needed to run a `Pipeline`. The `pipeline.dump()` method now contains a new key with the requirements, if any.

We can include requirements at the `Pipeline` level, and ideally we would add requirements for custom steps via
`@requirements` decorator, to avoid making the step definition more verbose.

It will throw a `ValueError` before running and show the dependencies that aren't already installed in your environment.

```python
@requirements(["distilabel>=0.0.1"])
class CustomStep(Step):
    @property
        return ["instruction"]

    @property
        return ["response"]

    def process(self, inputs: StepInput) -> StepOutput:  # type: ignore
        for input in inputs:
            input["response"] = "unit test"
        yield inputs

with BasePipeline(
    name="unit-test-pipeline", requirements=["random_requirement"]
) as pipeline:
    gen_step = DummyGeneratorStep()
    step1_0 = DummyStep()
    step2 = CustomStep()

    gen_step >> step1_0 >> step2
pipeline.run()
```


src/distilabel/utils/requirements.py
```diff
@@ -0,0 +1,57 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING, List, Union
+
+if TYPE_CHECKING:
+    from distilabel.steps import Step
+
+
+def add_requirements(requirements: Union[List[str]]) -> "Step":
```
gabrielmbmb, line 21: I would call this `requirements` instead of `add_requirements`, I think is more declarative.

gabrielmbmb: Also, the return type hint is wrong, it should be `Callable[["Step"], "Step"]`. Also, we should create a `TypeVar` bound to `_Step`. With the current setup the type of any class decorated will be coerced to `Step` (won't keep their original type)

gabrielmbmb (changes requested):
Looks good!

src/distilabel/mixins/requirements.py
```diff
@@ -0,0 +1,82 @@
...
+    def requirements(self) -> List[str]:
+        """Return a list of requirements that must be installed to run the `Pipeline`.
+
+        The requirements in a Pipeline will include the requirements from all the steps (if any).
+
+        Returns:
+            List of requirements that must be installed to run the `Pipeline`, sorted alphabetically.
+        """
+        self.requirements = self._gather_requirements()
+        return self._requirements
```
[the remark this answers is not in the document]
plaguss: In the way we intend to use it, this cannot happen, so I would go with the second option

docs/sections/how_to_guides/advanced/pipeline_requirements.md
```diff
@@ -0,0 +1,54 @@
+# Add requirements to run a Pipeline
+
+When sharing a pipeline that is not made of the predefined `Steps` and `Tasks` in distilabel, you may want to add the specific requirements that are needed to run it,
+to warn the user if something isn't installed and avoid errors when running.
```
gabrielmbmb, line 4:
```suggestion
When sharing a `Pipeline` that contains custom `Step`s or `Task`s, you may want to add the specific requirements that are needed to run them. `distilabel` will take this list of requirements and warn the user if any are missing.
```

docs/sections/how_to_guides/advanced/pipeline_requirements.md
```diff
@@ -0,0 +1,54 @@
...
+
+You can now add requirements at the step level as well as on the pipeline. The following is a sample pipeline that adds requirements `nltk` as a dependency to a sample `CustomStep`, and we also inform the pipeline must run using `distilabel>=1.3.0`:
```
gabrielmbmb, line 6:
```suggestion
Let's see how we can add additional requirements with an example. The first thing we're going to do is to add requirements for our `CustomStep`. To do so we use the `requirements` decorator to specify that the step has `nltk>=3.8` as dependency (we can use [version specifiers](https://peps.python.org/pep-0440/#version-specifiers)). In addition, we're going to specify at `Pipeline` level that we need `distilabel>=1.3.0` to run it.
```

docs/sections/how_to_guides/advanced/pipeline_requirements.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+with Pipeline(
+    name="pipeline-with-requirements", requirements=["distilabel>=1.3.0"]
+) as pipeline:
+    loader = LoadDataFromDicts(data=[{"instruction": "sample sentence"}])
+    step1 = CustomStep()
+    loader >> step1
+
+pipeline.run()
```
gabrielmbmb, line 42:
```suggestion

if __name__ == "__main__":
    pipeline.run()
```

docs/sections/how_to_guides/advanced/pipeline_requirements.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+Once we call `pipeline.run()`, if any of the requirements informed at the `Step` or `Pipeline` level isn't met, a `ValueError` will be raised telling us that we should install the list of dependencies:
```
gabrielmbmb, line 45:
```suggestion
Once we call `pipeline.run()`, if any of the requirements informed at the `Step` or `Pipeline` level is missing, a `ValueError` will be raised telling us that we should install the list of dependencies:
```

mkdocs.yml
```diff
@@ -162,6 +162,7 @@ nav:
           - Cache and recover pipeline executions: "sections/how_to_guides/advanced/caching.md"
           - Structured data generation: "sections/how_to_guides/advanced/structured_generation.md"
           - Serving an LLM for sharing it between several tasks: "sections/how_to_guides/advanced/serving_an_llm_for_reuse.md"
+          - Impose requirements to your pipelines and steps: "sections/how_to_guides/advanced/pipeline_requirements.md"
```
gabrielmbmb, line 165:
```suggestion
          - Specify requirements for pipelines and steps: "sections/how_to_guides/advanced/pipeline_requirements.md"
```

src/distilabel/pipeline/base.py
```diff
@@ -338,6 +343,13 @@ def run(
         # Load the `_BatchManager` from cache or create one from scratch
         self._load_batch_manager(use_cache)
 
+        if to_install := self.requirements_to_install():
+            # Print the list of requirements like they would appear in a requirements.txt
+            to_install_list = "\n" + "\n".join(to_install)
+            msg = f"Please install the following requirements to run the pipeline: {to_install_list}"
+            self._logger.error(msg)
+            raise ValueError(msg)
```
gabrielmbmb, line 351: maybe here is better a `ModuleNotFoundError`?

gabrielmbmb (approved):
LGTM!

PR #735, opened 2024-06-14
Add `MixtureOfAgentsLLM`

gabrielmbmb:
## Description

This PR adds the `MixtureOfAgents` llm, an special LLM that allows combining an `aggregator_llm` and a list of `proporsers_llms` as described in this paper: https://arxiv.org/abs/2406.04692. Although `LLM`s in `distilabel` are not in charge of tweaking the input, this special LLM does as it requires to inject the Mixture-of-Agents system prompt in the system prompt of the task allowing passing the references of the `proposers_llms`. 

> ⚠️ This system prompt injection should work fine with every task, although it hasn't been tested with all the tasks, only with the `TextGeneration`task.

In addition and motivated by the addition of the `MixtureOfAgentsLLM`, this PR adds/updates the following:

- The `RuntimeParametersMixin` has been updated to automatically handle attributes with type `List[RuntimeParametersMixin]`. The way to set the runtime parameters for the list of mixins is via a `dict` in which each key references the position of the mixin in the list to which the runtime parameters wants to be set.
- The CLI has been updated to be able to render `List[RuntimeParametersMixin]` attributes.
- Custom `from_dict` method of the `Step` has been removed as parent method is able to deserialise `_Serializable` attributes recursively.
- Remove duplicate `generate` method implementation of `AsyncLLM`s adding a new `_num_generations_param_supported ` flag that executes the right method to handle the `num_generations` parameter.


src/distilabel/pipeline/base.py
```diff
@@ -398,7 +398,9 @@ def dry_run(
         self._dry_run = False
         return distiset
 
-    def get_runtime_parameters_info(self) -> Dict[str, List[Dict[str, Any]]]:
+    def get_runtime_parameters_info(
+        self,
+    ) -> Dict[str, Union[List[Dict[str, Any]], Dict[str, Dict[str, Any]]]]:
```
plaguss, line 403: Wouldn't be more readable to have some alias for this kind of type hint? just a personal opinion

plaguss (approved):
LGTM!

src/distilabel/pipeline/base.py
gabrielmbmb: yes, I'll add it.

src/distilabel/llms/base.py
```diff
@@ -162,21 +172,27 @@ def get_runtime_parameters_info(self) -> List[Dict[str, Any]]:
...
         generation_kwargs_info = next(
-            runtime_parameter_info
-            for runtime_parameter_info in runtime_parameters_info
-            if runtime_parameter_info["name"] == "generation_kwargs"
+            (
+                runtime_parameter_info
+                for runtime_parameter_info in runtime_parameters_info
+                if runtime_parameter_info["name"] == "generation_kwargs"
+            ),
+            None,
```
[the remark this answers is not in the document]
gabrielmbmb: If we don't found the `runtime_parameter_info` with the `name == "generation_kwargs"`, then the generation expression `(... for item in list if item == "something")` will be empty and `next` will raise an `StopGeneration` exception. To avoid that, we set `None` as default for that cases.

PR #741, opened 2024-06-18
Add `examples/arena_hard.py` and remove from `distilabel` core

alvarobartt:
## Description

As recently discussed internally, maintaining the benchmarks may be too time consuming and hard, since there are a lot of things going on and benchmarks tend to be maintained separately. So on, we internally decided to remove the implementation from the `distilabel` core for the moment, and provide that as an example instead, that anyone can use as reference for building benchmarks with `distilabel`.


gabrielmbmb (approved):
🪦

PR #745, opened 2024-06-20
Fix docs for distiset.save_to_disk kwargs

fpreiss:
The keyword argument for the path in Distiset.save_to_disk should be documented as `distiset_path` instead of `dataset_path`.


plaguss (approved):
Thanks for the catch @fpreiss! 😃

PR #747, opened 2024-06-21
Add new step `CombineKeys`

plaguss:
## Description

This PR adds a new step to combine keys in a dict. The main use case I found was the following:

In a `Pipeline` I have a `Task` that generates instructions (say `GenerateSentencePair`), and I want more examples of the `positives` generated for diversity. With a new task, I generate more of those, so afterwards I want to combine the `positive` keys with the new extra instructions (say a list of instructions called `positives` or whatever). With this `CombineKeys` we can combine those keys:

```python
from distilabel.steps import CombineKeys

combiner = CombineKeys(
    keys=["queries", "multiple_queries"],
    output_key="queries",
)
combiner.load()

result = next(
    combiner.process(
        [
            {
                "queries": "How are you?",
                "multiple_queries": ["What's up?", "Everything ok?"]
            }
        ],
    )
)
# >>> result
# [{'queries': ['How are you?', "What's up?", 'Everything ok?']}]
```

### DIFF
--- a/src/distilabel/steps/combine.py
+++ b/src/distilabel/steps/combine.py
@@ -124,7 +124,7 @@ class CombineColumns(Step):
 
 
 class CombineKeys(Step):
-    """Combines keys from a `StepInput`.
+    """Combines keys from a row.
 
     `CombineKeys` is a `Step` that implements the `process` method that calls the `combine_keys`
     function to handle and combine keys in a `StepInput`. `CombineKeys` provides two attributes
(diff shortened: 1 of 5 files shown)


src/distilabel/steps/combine.py
```diff
@@ -121,3 +121,80 @@ def process(self, *inputs: StepInput) -> "StepOutput":
             merge_keys=self.inputs,
             output_merge_keys=self.outputs,
         )
+
+
+class CombineKeys(Step):
+    """Combines keys from a `StepInput`.
```
gabrielmbmb, line 127: As this short description is mainly for the component gallery and users don't necessarily know the `StepInput` I would change it to "Combine keys from a row"

gabrielmbmb (approved):
LGTM!

Issue #756, opened 2024-06-26
[FEATURE] remove deprecated `CombineColumns` 


davidberenstein1957:
**Is your feature request related to a problem? Please describe.**
Within 1.3 we deprecated this naming as part of #755.

**Describe the solution you'd like**
N.A.

**Describe alternatives you've considered**
N.A.

**Additional context**
N.A.

PR #758, opened 2024-06-26
Refactor naming columns steps combinecolumns combinekeys expandcolumns

### LINKED ISSUE #755: [FEATURE] review naming columns `Steps`: `CombineColumns`, `CombineKeys`, `ExpandColumns`
The naming for these operations is a bit fragmented and not always as explicit/intuitive.

I propose reviewing the naming and perhaps aligning it with a more familiar naming and API design like the one used in pandas.

- `ExpandColumns` is nice and explicit, but I think pandas uses explode, so something like `ExplodeColumns` would perhaps be better.
- `CombineKeys` uses the Keys in the name, but I think it would be better to go for something like `AggregateColumns`/ `MergeColumns`, `AppendColumns`  
- `CombineColumns` seems to groupby operation and not really a combination. Perhaps something like `GroupByColumns` might be better.

Open to iterating on the naming.

N.A.
**Additional context**
N.A.

davidberenstein1957:
Closes #755 

- Renamed `CombineKeys` -> `MergeColumns`
- Renamed `CombineColumns` -> `GroupColumns`
- Moved column operations to `distilabel/steps/columns/*.py`
- Added #756  w.r.t. deprecation


src/distilabel/pipeline/utils.py
```diff
@@ -52,32 +54,34 @@ def combine_dicts(
...
                     combined_dict[key] = value
         result.append(combined_dict)
     return result
 
 
-def combine_keys(row: Dict[str, Any], keys: List[str], new_key: str = "combined_key"):
-    """Combines keys in a dictionary into a single key on the specified `new_key`.
+def merge_columns(
+    row: Dict[str, Any], columns: List[str], new_column: str = "combined_key"
+):
```
gabrielmbmb, line 68:
```suggestion
) -> Dict[str, Any]:
```

PR #778, opened 2024-07-10
Add `Magpie` and `MagpieGenerator` tasks

gabrielmbmb:
## Description

This PR adds two new tasks `Magpie` and `MagpieGenerator` based on [Magpie: Alignment Data Synthesis from Scratch by Prompting Aligned LLMs with Nothing](https://arxiv.org/abs/2406.08464) paper, which presents a new way to leverage chat template and the autoregressive capabilities of the instruct fine-tuned LLMs to generate instructions without seed data or a specific system prompt.

In addition, a new mixin `MagpieChatTemplateMixin` has been added that has to be used in combination with the aforementioned tasks. This mixin requires `LLM`s that are able to send the raw input (with special tokens, etc) to the LLM, so at this moment only the `TransformersLLM`, `vLLM` and `InferenceEndpointsLLM` uses the mixin and therefore are the only LLMs that can be used with the new tasks.

Finally, `InferenceEndpointsLLM` has been updated and the `use_openai_client` argument has been removed (and the use of the OpenAI client) as it was not used and didn't add any additional functionality that the `huggingface_hub.AsyncInferenceClient` doesn't offer. In addition, if `tokenizer_id` is not provided, then the `chat_completion` method from the `huggingface_hub.AsyncInferenceClient` is used instead of the `text_generation`. ⚠️ Even though with the new `chat_generation` methods a list of `tools` can be passed, we need to update the `LLM.generate` function signature and adapt  the `agenerate` method so function calls can be returned too #782 (I'll tackle this in a different PR).


src/distilabel/steps/tasks/magpie/base.py
```diff
@@ -0,0 +1,267 @@
...
+from distilabel.steps.base import StepInput
+from distilabel.steps.tasks.base import Task
+
+if TYPE_CHECKING:
+    from distilabel.steps.tasks.typing import ChatType, FormattedInput
+    from distilabel.steps.typing import StepOutput
+
+MAGPIE_MULTI_TURN_SYSTEM_PROMPT = (
+    "You are a helpful Al assistant. The user will engage in a multi−round conversation"
+    " with you,asking initial questions and following up with additional related questions."
```
plaguss, line 34:
```suggestion
    " with you, asking initial questions and following up with additional related questions."
```

src/distilabel/steps/tasks/magpie/base.py
```diff
@@ -0,0 +1,267 @@
...
+)
+from distilabel.steps.base import StepInput
+from distilabel.steps.tasks.base import Task
+
+if TYPE_CHECKING:
+    from distilabel.steps.tasks.typing import ChatType, FormattedInput
+    from distilabel.steps.typing import StepOutput
+
+MAGPIE_MULTI_TURN_SYSTEM_PROMPT = (
+    "You are a helpful Al assistant. The user will engage in a multi−round conversation"
```
plaguss, line 33: Maybe this should be changed?
```suggestion
    "You are a helpful Al assistant. The user will engage in a multi−turn conversation"
```

src/distilabel/steps/tasks/magpie/base.py
```diff
@@ -0,0 +1,267 @@
...
+
+class Magpie(Task, MagpieBase):
+    """Generates conversations using an instruct fine-tuned LLM.
+
+    Magpie is a neat method that allows generating user instructions with no seed data
+    or specific system prompt thanks to the autoregressive capabilities of the instruct
+    fine-tuned LLMs. As they were fine-tuned using a chat template composed by a user message
+    and a desired assistant output, the instruct fine-tuned LLM learns that after the pre-query
+    or pre-instruct tokens comes an instruction. If these pre-query tokens are sent to the
+    LLM without any user message, then the LLM will continue generating tokens as it was
```
plaguss, line 187:
```suggestion
    LLM without any user message, then the LLM will continue generating tokens as if it was
```

plaguss (commented):
LGTM! will review again once it's ready for review 👍

src/distilabel/steps/tasks/magpie/base.py
gabrielmbmb: I copy pasted it from the paper 😋

plaguss (approved):

plaguss (approved):

PR #805, opened 2024-07-22
Compatibility with `vLLM` with `tensor_parallel_size` argument

gabrielmbmb:
## Description

This PR adds a few changes that enables using `vLLM` with the argument [`tensor_parallel_size`](https://docs.vllm.ai/en/latest/serving/distributed_serving.html). This argument enables the use of multiple GPUs with vLLM, and to do so, vLLM uses `multiprocessing` or `ray`:

1. When using the `multiprocessing` approach, it was not working as `vLLM` was trying to create new processes but it was not able to because the process that `distilabel` creates was a daemon process, which is not allowed to create child processes. To by pass these issue, a `_NoDaemonPool` class has been created that creates non-daemon processes and it's used in `Pipeline`.
2. With `ray` it works installing the version in the `main` branch which includes the changes of https://github.com/vllm-project/vllm/pull/6032. It's needed to set `VLLM_USE_RAY_COMPILED_DAG=1` and `VLLM_USE_RAY_SPMD_WORKER=1` environment variables. **Update**: `vllm==0.5.3` has been released which includes the changes needed.

### DIFF
--- a/src/distilabel/pipeline/local.py
+++ b/src/distilabel/pipeline/local.py
@@ -69,10 +69,10 @@ class _NoDaemonContext(type(mp.get_context())):
 class _NoDaemonPool(Pool):
     def __init__(
         self,
-        processes: int | None = None,
-        initializer: Callable[..., object] | None = None,
+        processes: Union[int, None] = None,
+        initializer: Union[Callable[..., object], None] = None,
         initargs: Iterable[Any] = ...,  # type: ignore
-        maxtasksperchild: int | None = None,
+        maxtasksperchild: Union[int, None] = None,
     ) -> None:
         super().__init__(
             processes=processes,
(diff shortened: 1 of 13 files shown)


src/distilabel/pipeline/local.py
```diff
@@ -48,6 +49,40 @@ def _init_worker(log_queue: "Queue[Any]") -> None:
     setup_logging(log_queue)
 
 
+# We create a custom `Pool` class so the created processes are not daemons, allowing
+# them to created child processes if necessary (for example when using `vLLM` with `tensor_parallel_size`)
```
plaguss, line 53:
```suggestion
# them to create child processes if necessary (for example when using `vLLM` with `tensor_parallel_size`)
```

plaguss (approved):

PR #809, opened 2024-07-23
Add default name for a pipeline

plaguss:
## Description

Generate a random name for the pipeline if one is not given by default, to simplify cognitive load:

```python

with Pipeline() as pipeline:
    print(pipeline.name)
```

A name like the `pipeline_33prTGVu` will appear by default from now on, unless the name is given.


src/distilabel/pipeline/base.py
```diff
@@ -122,6 +124,17 @@ def get_pipeline(cls) -> Union["BasePipeline", None]:
...
 
 
+def generate_random_name(num_chars: int = 8) -> str:
+    """Helper function to generate a random name for a pipeline by default.
+
+    Returns:
+        Random name for a pipeline.
+    """
+    return "pipeline_" + "".join(
+        random.choice(string.ascii_letters + string.digits) for _ in range(num_chars)
```
davidberenstein1957, line 134: why not use something like UUID?

plaguss: no real reason, I've just updated it as it's less code to maintain, thanks 😄

davidberenstein1957: Exactly my thoughts :)

plaguss:
Hi @davidberenstein1957, the LLMs don't need it, the steps already have it implemented

davidberenstein1957 (approved):

davidberenstein1957:
@plaguss perhaps we should update the docs where needed?

plaguss:
Yes totally, there are some updates in  https://github.com/argilla-io/distilabel/pull/812 to the quickstart for example that should contain this (it should be here, but they are a bit mixed). Maybe we should review the docs after these 2 PRs are merged and see what to improve? Thanks for the review 🤗

PR #837, opened 2024-07-29
Refactor 836 refactor convert inputs and outputs to class property 

### LINKED ISSUE #836: [REFACTOR] convert `inputs` and `outputs` to class property
I cannot access the input and outputs of a class without initializing it.

Given that they are fixed, I would say something like the code below might be nicer and allow accessing without initialisation and we can reduce the code-base a bit, not having to define the property methods for each class and just inherit them from the parent Step/Task.

OLD: 

```python
from typing import List

class Task:
    pass

class TextGeneration(Task):
    use_system_prompt: bool = True

    @property
    def inputs(cls) -> List[str]:
        """The input for the task is the `instruction`."""
        return ["instruction"]

    @property
    def outputs(cls) -> List[str]:
        """The input for the task is the `instruction`."""
        return ["response"]

# Accessing the property directly from the class
print(TextGeneration._inputs)  # Output: ["instruction"]
```

Remove this from each child class and 

```python
@property
    """The input for the task is the `instruction`."""
    return ["instruction"]

@property
    """The input for the task is the `instruction`."""
    return ["response"]
```

Rewrite parent classes like `Step` etc.
```python
@property
    """The input for the task is the `instruction`."""
    return self._inputs

@property
    """The input for the task is the `instruction`."""
    return self._outputs
```

New class defintition
```python
from typing import List

class Task:
    pass

class TextGeneration(Task):
    use_system_prompt: bool = True

    _inputs = ["instruction"]
    _outputs = ["response"]

print(TextGeneration._inputs)  # Output: ["instruction"]
print(TextGeneration(**init_kwargs).inputs)  # Output: ["instruction"]

```

NA

**Additional context**
NA

### LINKED ISSUE #839: [REFACTOR] unify singular/plural semantic naming of columns
I feel distinguishing between singular and plural names complicates workflows and code, e.g., “generation” and “generations”, and I’ve seen similar scenarios with other columns like “answer”/“rationale”.  I’ve seen similar scenarios with other columns like “answer”/“rationale”.

Let's simplify this and keep only the plural

NA

**Additional context**
NA

davidberenstein1957:
Closes #836 
Closes #839


src/distilabel/steps/tasks/magpie/generator.py
```diff
@@ -209,35 +208,6 @@ class MagpieGenerator(GeneratorTask, MagpieBase):
...
-        """Checks that the provided `LLM` uses the `MagpieChatTemplateMixin`."""
-        super().model_post_init(__context)
-
-        if not isinstance(self.llm, MagpieChatTemplateMixin):
-            raise ValueError(
-                f"`Magpie` task can only be used with an `LLM` that uses the `MagpieChatTemplateMixin`."
-                f"`{self.llm.__class__.__name__}` doesn't use the aforementioned mixin."
-            )
-
-        self.llm.use_magpie_template = True
```
davidberenstein1957, line 222: @gabrielmbmb was there a reason you were defining these again? They were also defined in the parent class.

gabrielmbmb: Yes, having the `model_post_init` in the `MagpieBase` class and then inheriting in `Magpie` and `MagpieGenerator` didn't trigger it, so that's why it's duplicate in `Magpie` and `MagpieGenerator` tasks. Apparently, this is an issue with pydantic that should be fixed but it's not working and didn't want to expend too much time so I decided to duplicate the code.

PR #858, opened 2024-08-06
Add OpenAI request timeout

ashim-mahara:
The timeout kwarg for OpenAILLM was missing causing in timeouts when using OpenAI compatible servers that might take time to respond to requests.


plaguss (approved):
Thanks for the contribution!

plaguss:
After reviewing with @gabrielmbmb we noticed this parameter can be informed at the client level:

```python
from distilabel.llms import OpenAILLM

oai_llm = OpenAILLM(..., timeout=123)
```

So there's no need to pass the parameter again as a runtime parameter, but just to remove the internal timeout that was overriding the client definition.

PR #868, opened 2024-08-08
Add default structured output for `GenerateSentencePair` task

### LINKED ISSUE #867: [FEATURE] `GenerateSentencePair` predefined structured output

plaguss:
## Description

Adds a default structured output for `GenerateSentencePair`, it can be set from the `Task` and will prepare the LLM for it internally.

```python
from distilabel.steps.tasks import GenerateSentencePair
from distilabel.llms import InferenceEndpointsLLM

generate_sentence_pair = GenerateSentencePair(
    triplet=True, # `False` to generate only positive
    action="query",
    llm=...,
#    llm=InferenceEndpointsLLM(
#        model_id="meta-llama/Meta-Llama-3.1-70B-Instruct",
#    ),
    use_default_structured_output=True,
    input_batch_size=10,
)
```

Depending of the `LLM` will use a provider for the structured outputs. For example with `InferenceEndpointsLLM` it will use `outlines` to generate a dict of the form: `{"positive": ...}` or `{"positive": ..., "negative": ...}` that will be easier to parse.

It also puts in place a mechanism to define a default structured output for every `Task` that may require it.

Closes #867

### DIFF
--- a/src/distilabel/steps/tasks/base.py
+++ b/src/distilabel/steps/tasks/base.py
@@ -155,14 +155,25 @@ class _Task(_Step, ABC):
         return output
 
     def _set_default_structured_output(self) -> None:
-        """Prepares the structured output to be set in the selected `LLM`."""
+        """Prepares the structured output to be set in the selected `LLM`.
+
+        If the method `get_structured_output` returns None (the default), there's no need
+        to set anything, as it doesn't apply.
+        If the `use_default_structured_output` and there's no previous structured output
+        set by hand, then decide the type of structured output to select depending on the
+        `LLM` provider.
+        """
+        schema = self.get_structured_output()
+        if not schema:
+            return
+
         if self.use_default_structured_output and not self.llm.structured_output:
             # In case the default structured output is required, we have to set it before
             # the LLM is loaded
             from distilabel.llms import InferenceEndpointsLLM
             from distilabel.llms.base import AsyncLLM
 
-            structured_output = {"schema": self.get_structured_output()}
+            structured_output = {"schema": schema}
             # To determine instructor or outlines format
             if not (
                 isinstance(self.llm, AsyncLLM)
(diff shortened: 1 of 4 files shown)


src/distilabel/steps/tasks/sentence_transformers.py
```diff
@@ -337,3 +364,57 @@ def format_output(
...
+                "type": "object",
+            }
+
+        # To determine instructor or outlines format
+        if isinstance(self.llm, AsyncLLM) and not isinstance(
+            self.llm, InferenceEndpointsLLM
+        ):
+            return {"schema": schema}
+
+        return {"format": "json", "schema": schema}
```
gabrielmbmb, line 403: We should move this logic so it doesn't need to be coded in every task. Maybe we can just return the `schema` in the `get_structured_output` method, and then move this condition to the `Task.load` method.

src/distilabel/steps/tasks/base.py
```diff
@@ -63,10 +63,16 @@ class _Task(_Step, ABC):
...
     )
+    use_default_structured_output: bool = False
 
     def load(self) -> None:
         """Loads the LLM via the `LLM.load()` method."""
         super().load()
+        if self.use_default_structured_output and not self.llm.structured_output:
+            # In case the default structured output is required, we have to set it before
+            # the LLM is loaded
+            self.llm.structured_output = self.get_structured_output()
```
gabrielmbmb, line 74: maybe we can move this to a method

src/distilabel/steps/tasks/base.py
```diff
@@ -63,10 +63,16 @@ class _Task(_Step, ABC):
     num_generations: RuntimeParameter[int] = Field(
         default=1, description="The number of generations to be produced per input."
     )
+    use_default_structured_output: bool = False
```
gabrielmbmb, line 66: I think this can have as default value `True`

gabrielmbmb (approved):
LGTM!

PR #871, opened 2024-08-08
Add `save_artifact` method to `_Step`

### LINKED ISSUE #834: [FEATURE] Allow `Step`s to produce artifacts
Some `Step`s might produce some artifacts (faiss index for example) that could be reused after the pipeline has finished its execution. `distilabel` currently doesn't allow to save any artifact, or does not provide a clear way, for an `Step` to save artifacts.

`Step`s can produce and save artifacts that can be later upload along the produced dataset to the Hub. To do so, `Step` offers a method `get_artifact_directory` that can be used within the step to a get directory (from the cache of the pipeline) that can be used by the step to store produced artifacts.

gabrielmbmb:
## Description

This PR adds a new method called `save_artifact` to the `_Step` base class that allows saving artifacts generated by an `Step` such as the Faiss index generated by `FaissNearestNeighbour` step. The generated artifacts will be uploaded to the Hugging Face Hub when using the `push_to_hub` method of the `Distiset` class.

Closes #834


src/distilabel/distiset.py
```diff
@@ -425,26 +481,31 @@ def load_from_disk(
...
+            elif folder.stem == DISTISET_ARTIFACTS_FOLDER:
+                has_artifacts = True
+                continue
             distiset[folder.stem] = load_from_disk(
                 str(folder),
                 keep_in_memory=keep_in_memory,
             )
-        # From the config folder we just need to point to the files. Once downloaded we set the path
+
+        # From the config folder we just need to point to the files. Once downloaded we set the pathto pointto point to the files. Once downloaded we set the path
```
plaguss, line 508:
```suggestion
        # From the config folder we just need to point to the files. Once downloaded we set the path to point to point to the files. Once downloaded we set the path
```

src/distilabel/utils/card/distilabel_template.md
```diff
@@ -70,6 +70,21 @@ ds = load_dataset("{{ repo_id }}")
...
 
+{% if artifacts %}
+## Artifacts
+
+{% for step_name, artifacts in artifacts.items() %}
+* **Step**: `{{ step_name }}`
+  {% for artifact in artifacts %}
+    * **Artifact name**: `{{ artifact.name }}`
+      {% for name, value in artifact.metadata.items() %}
+        * `{{ name }}`: {{ value }}
```
plaguss, line 81: The `value` of the artifact's metadata here points to the object stored in the repo? I'm not sure, maybe I missed in the code, but would be helpful to easily point to it

plaguss (approved):
LGMT! Maybe we should add docs for this

src/distilabel/utils/card/distilabel_template.md
gabrielmbmb: Renders the values in the `metadata.json` of each artefact, is that what you were asking?

PR #877, opened 2024-08-09
Fix loader to read from a glob pattern

plaguss:
## Description

The data loader `LoadDataFromFileSystem` couldn't read from a glob pattern, no it's solved:

```python
from distilabel.steps import LoadDataFromFileSystem

loader = LoadDataFromFileSystem(
    data_files="path/to/dataset/*.jsonl",
    streaming=True
)
loader.load()

result = next(loader.process())
# >>> result
# ([{'type': 'function', 'function':...', False)
```

Using this dataloader form we can read from a folder made up of a lot of files with lots of examples (thinking of fine web).


src/distilabel/steps/generators/huggingface.py
```diff
@@ -387,6 +405,11 @@ def get_filetype(data_path: UPath) -> str:
         if data_path.is_file():
             filetype = get_filetype(data_path)
             data_files = str(data_path)
+
+        elif len(glob.glob(str(data_path))) >= 1:
```
gabrielmbmb, line 409: I think `glob` package won't work with `s3://...` or other remote file systems. I suggests the following:

```python
parent = data_path.parent
len(parent.glob(data_path.name))
```

plaguss: my bad!

PR #878, opened 2024-08-09
Generate deterministic pipeline name when it's not given

### LINKED ISSUE #875: [BUG] The default random name of a pipeline prevents from using the cache
**Describe the bug**
To simplify the code, if no name is given to a pipeline a random one is generated.

**To Reproduce**
Code to reproduce
```python

with Pipeline() as pipe:
    ...
```

**Expected behaviour**
Make the default name deterministic

**Screenshots**

**Desktop (please complete the following information):**
 - Package version:
 - Python version:

**Additional context**

plaguss:
## Description

When generating a name for the pipeline automatically we were doing it with a random generator, which prevents the cache from working. This changes the behaviour to be deterministic, using the signature of the pipeline (before running, but after closing the context manager so the steps are taken into account even if not all the information is available at that point).

*Note*: Additionally, a test was failing because it wasn't refreshing the cache folder between tests.

Closes #875


src/distilabel/pipeline/base.py
```diff
@@ -231,6 +232,8 @@ def __enter__(self) -> Self:
     def __exit__(self, exc_type, exc_value, traceback) -> None:
         """Unset the global pipeline instance when exiting a pipeline context."""
         _GlobalPipelineManager.set_pipeline(None)
+        if self.name == _PIPELINE_DEFAULT_NAME:
+            self.name = f"pipeline_{self._create_signature()[:8]}"
```
gabrielmbmb, line 236: what if instead of using `self._create_signature()`, we append the names of the step in the pipeline? With this solution, changing one parameter of one step will create another directory.

gabrielmbmb (changes requested):

src/distilabel/pipeline/base.py
plaguss: Sure, makes more sense

gabrielmbmb (approved):

PR #883, opened 2024-08-12
Update `inputs` and `outputs` interface to allow returning dict indicating optionality 

gabrielmbmb:
## Description

This PR updates the method signature of `inputs` and `outputs` to also allow returning a dictionary in which the keys are the name of the columns and the values if they are required or not. This change is required as some `Step`s have optionals columns (meaning that they return an `[]` empty list), but the internal logic as it is doesn't allow to use `input_mappings` and `output_mappings` with these optional columns.


plaguss (approved):
LGTM

Issue #880, opened 2024-08-12
[FEATURE] Add `exclude_from_signature` attribute in `Step`


gabrielmbmb:
**Is your feature request related to a problem? Please describe.**
As all the attributes of an `Step` are used to compute the signature of the pipeline, changing one attribute that shouldn't affect on the results, such as the `input_batch_size`, `gpu_memory_utilization`, `resources`, etc change the signature of the pipeline not allowing to reuse the cache of previous executions with the same generation parameters.

**Describe the solution you'd like**
Add an attribute called `exclude_from_signature` (via mixin?) that allows specifying the names of attributes of the class that shouldn't be used to compute the signature.

PR #894, opened 2024-08-13
Move `_create_signature` to mixin and add new `exclude_from_signature` argument

plaguss:
## Description

This PR adds a new mixin that contains the `_create_signature` method for the pipeline, and adds a new attribute named `exclude_from_signature` which consists of a `set` of strings that won't be taken into account when creating the signature for the pipeline.

Closes https://github.com/argilla-io/distilabel/issues/880


plaguss:
> Maybe it's better to merge this one after #766? I think you will have to change a few things

Yes, will let it rest until that one is done

PR #902, opened 2024-08-14
New `TruncateTextColumn` to truncate the length of texts using the number of tokens or characters

plaguss:
## Description

This PR adds a new `TruncateTextColumn` step to truncate the text.

While developing pipelines that start from datasets with longer texts, the model can fail for longer texts than the max context length the model can handle. This step can be useful to avoid such errors:

- Truncating a row to a given number of tokens:

```python
from distilabel.steps import TruncateTextColumn

trunc = TruncateTextColumn(
    tokenizer="meta-llama/Meta-Llama-3.1-70B-Instruct",
    max_length=4,
    column="text"
)

trunc.load()

result = next(
    trunc.process(
        [
            {"text": "This is a sample text that is longer than 10 characters"}
        ]
    )
)
# result
# [{'text': 'This is a sample'}]
```

- Truncating a row to a given number of characters

```python
from distilabel.steps import TruncateTextColumn

trunc = TruncateTextColumn(max_length=10)

trunc.load()

result = next(
    trunc.process(
        [
            {"text": "This is a sample text that is longer than 10 characters"}
        ]
    )
)
# result
# [{'text': 'This is a '}]
```

### DIFF
--- a/src/distilabel/steps/__init__.py
+++ b/src/distilabel/steps/__init__.py
@@ -47,6 +47,7 @@ from distilabel.steps.generators.huggingface import (
 from distilabel.steps.generators.utils import make_generator_step
 from distilabel.steps.globals.huggingface import PushToHub
 from distilabel.steps.reward_model import RewardModelScore
+from distilabel.steps.truncate import TruncateRow
 from distilabel.steps.typing import GeneratorStepOutput, StepOutput
 
 __all__ = [
@@ -80,4 +81,5 @@ __all__ = [
     "GeneratorStepOutput",
     "StepOutput",
     "step",
+    "TruncateRow",
 ]
--- a/src/distilabel/steps/__init__.py
+++ b/src/distilabel/steps/__init__.py
@@ -78,8 +78,8 @@ __all__ = [
     "Step",
     "StepInput",
     "RewardModelScore",
+    "TruncateRow",
     "GeneratorStepOutput",
     "StepOutput",
     "step",
-    "TruncateRow",
 ]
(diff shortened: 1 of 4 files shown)


src/distilabel/steps/__init__.py
```diff
@@ -80,4 +81,5 @@
     "GeneratorStepOutput",
     "StepOutput",
     "step",
+    "TruncateRow",
```
gabrielmbmb, line 84:
```suggestion
    "TruncateRow",
    "GeneratorStepOutput",
    "StepOutput",
    "step",
```

src/distilabel/steps/truncate.py
```diff
@@ -0,0 +1,149 @@
...
+            )
+        )
+        # result
+        # [{'text': 'This is a '}]
+        ```
+    """
+
+    column: str = "text"
+    max_length: int = 8192
+    tokenizer_name: Optional[str] = None
```
gabrielmbmb, line 102: I think it's better just to name this attribute `tokenizer` to keep consistency with the other steps, llms, etc.

src/distilabel/steps/truncate.py
```diff
@@ -0,0 +1,149 @@
...
+    _truncator: Optional[Callable[[str], str]] = None
+    _tokenizer: Optional[Any] = None
+
+    def load(self):
+        super().load()
+        if self.tokenizer_name:
+            if not importlib.util.find_spec("transformers"):
+                raise ImportError(
+                    "transformers is needed to tokenize, but is not installed. "
```
gabrielmbmb, line 111:
```suggestion
                    "`transformers` is needed to tokenize, but is not installed. "
```

gabrielmbmb (approved):

gabrielmbmb:
Maybe `TruncateTextColumn` is a better name for this step?

PR #910, opened 2024-08-19
Fix load data from disk

### LINKED ISSUE #909: How to load local hf dataset？
Hi, I have run the example provided on this [page](https://github.com/argilla-io/distilabel?tab=readme-ov-file#example) and it run well. Then I tried to load a local HF dataset and found that the class LoadDataFromDisk. 

I saved the dataset using the `save_to_disk` method like `dataset.save_to_disk(dataset_dict_path='./save_data') ` and following is the structure of the saved dataset directory.

save_data/
├── dataset_dict.json
└── test
    ├── data-00000-of-00001.arrow
    ├── dataset_info.json
    └── state.json

I only changed the "repo_id" to "dataset_path" (a local LLM is used). When I run the following code, a Value error encountered 
```
ValueError: Step 'load_data_from_disk_0' is missing required runtime parameter 'repo_id'. Please, provide a value for it when calling `Pipeline.run` method:

    pipeline.run(parameters={"load_data_from_disk_0": {"repo_id": ...}})
```

```
from distilabel.llms import vLLM
from distilabel.steps import LoadDataFromDisk
from distilabel.steps.tasks import TextGeneration

with Pipeline(
    name="simple-text-generation-pipeline",
    description="A simple text generation pipeline",
) as pipeline:
    load_dataset = LoadDataFromDisk(output_mappings={"prompt": "instruction"})

    generate_with_openai = TextGeneration(llm=vLLM(model="/home/public_data/qwen/Qwen2-7B-Instruct/"))

    load_dataset >> generate_with_openai

if __name__ == "__main__":
    distiset = pipeline.run(
        parameters={
            load_dataset.name: {
                "dataset_path": "/home/qiey/DataGenration/save_data",
                "split": "test",
            },
            generate_with_openai.name: {
                "llm": {
                    "generation_kwargs": {
                        "temperature": 0.7,
                        "max_new_tokens": 512,
                    }
                }
            },
        },
    )

```
could anyone give me some advice, any suggestion is appreciated.

plaguss:
## Description

This PR makes the `repo_id` argument in `LoadFromDisk` optional, and provides sets `config="default"` to simplify loading data if possible.

Closes #909


src/distilabel/steps/generators/huggingface.py
```diff
@@ -567,8 +570,13 @@ def load(self) -> None:
...
                 storage_options=self.storage_options,
             )
-            if self.config:
-                ds = ds[self.config]
+            if self.config not in ds.keys():
+                # TODO: Add FAQ for this case, pointing to the Distiset documentation on configurations.
+                raise ValueError(
+                    f"Configuration '{self.config}' not found in the Distiset, available ones"
+                    f" are: {list(ds.keys())}."
+                )
```
gabrielmbmb, line 578: Maybe we can add an additional phrase guiding the users on what they have to do to potentially fix the issue:

"... Please, try changing `config` attribute/runtime parameter value to one of the valid configurations"

src/distilabel/steps/generators/huggingface.py
```diff
@@ -597,8 +605,6 @@ def outputs(self) -> List[str]:
         """
         # We assume there are Dataset/IterableDataset, not it's ...Dict counterparts
         if self._dataset is Ellipsis:
```
gabrielmbmb, line 607:
```suggestion
        if self._dataset is None:
```

gabrielmbmb (changes requested):

src/distilabel/steps/generators/huggingface.py
plaguss: I've updated the error message to be more informative, but this is a candidate for a refactor after https://github.com/argilla-io/distilabel/pull/911

gabrielmbmb (approved):

PR #911, opened 2024-08-19
Add custom errors

plaguss:
## Description

This adds a new module to `distilabel` to store custom errors. The idea is to raise more informative errors (meaning, we point to the docs when possible).

Ideas for improvement are welcome. It's based on the treatment that pydantic does to the errors: https://github.com/pydantic/pydantic/blob/main/pydantic/errors.py. We don't have a specific page for errors, but instead we can point to whatever place that can solve the issue.

Currently `DistilabelUserError` (for `ValueError` exceptions) and `DistilabelTypeError` (for `TypeError` exceptions) have been added, which are just the base error type but the message allows passing a reference to a page in our docs, so we can show it when an error raises to redirect the user to the page:

Previously:

```python
>>> raise ValueError("This is an error message.")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: This is an error message.
```

Now (when we have information documented):

```python
>>> raise DistilabelUserError(
...     "This is an error message.", page="sections/getting_started/faq/"
... )
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>:1                                                                                    │
│ ╭─────────────────────────────── locals ────────────────────────────────╮                        │
│ │ DistilabelUserError = <class 'distilabel.errors.DistilabelUserError'> │                        │
│ ╰───────────────────────────────────────────────────────────────────────╯                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
DistilabelUserError: This is an error message.

For further information visit https://distilabel.argilla.io/latest/sections/getting_started/faq/
```


src/distilabel/errors.py
```diff
@@ -0,0 +1,61 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Final, Optional
+
+DISTILABEL_DOCS_URL: Final[str] = "https://distilabel.argilla.io/latest/"
```
gabrielmbmb, line 17: Should we move this to `constants.py`?

src/distilabel/errors.py
```diff
@@ -0,0 +1,61 @@
...
+# The sitemap can be visited for the full list of pages:
+# SITEMAP_URL: Final[str] = "https://distilabel.argilla.io/latest/sitemap.xml"
+
+
+class DistilabelErrorMixin:
```
gabrielmbmb, line 22: I would remove the `Mixin` suffix as it's more a base class than a mixin.

src/distilabel/errors.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __init__(self, message: str, *, page: Optional[str] = None) -> None:
+        self.message = message
+        self.page = page
+
+    def __str__(self) -> str:
+        if self.page is None:
+            return self.message
+        else:
+            return f"{self.message}\n\nFor further information visit {DISTILABEL_DOCS_URL}{self.page}"
```
gabrielmbmb, line 49:
```suggestion
            return f"{self.message}\n\nFor further information visit '{DISTILABEL_DOCS_URL}{self.page}'"
```

gabrielmbmb (approved):
LGTM!

PR #912, opened 2024-08-19
Fix `StructuredGeneration` examples and internal check

### LINKED ISSUE #906: [FEATURE] Unify API and documentation of structured generation
The current structured generation dependent on the framework we use underneath to generate structured outputs.

a single way to generate structured outputs that doesn't depend on the framework we use, it is confusing

**Additional context**

plaguss:
## Description

The examples of `StructuredGeneration` have been updated to show how the values can be informed, they are the same as a `TextGeneration` that takes a `structured_output` argument.

Both `AsyncLLM` and `LLM` (plus `InferenceEndpointsLLM`) take the same input:

```python
{"format": "json", "schema": "..."}
```

If the `format` is not informed (outlines requires that), we will try to infer it from the `schema`.

Closes #906

### DIFF
--- a/src/distilabel/llms/base.py
+++ b/src/distilabel/llms/base.py
@@ -428,7 +428,10 @@ class AsyncLLM(LLM):
         # We can deal with json schema or BaseModel, but we need to convert it to a BaseModel
         # for the Instructor client.
         schema = structured_output.get("schema", {})
-        if not issubclass(schema, BaseModel):
+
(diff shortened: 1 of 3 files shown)


davidberenstein1957:
@plaguss can ytou add the index page for deepseek prover here too? it does not show up at the moment. https://distilabel.argilla.io/dev/sections/pipeline_samples/examples/#examples

gabrielmbmb (approved):

PR #916, opened 2024-08-20
feat: 810 feature add huggingface hubutilstelemetry

### LINKED ISSUE #810: [FEATURE] Add `huggingface_hub.utils.telemetry`
We want to include some basic telemetry to understand the usage of distilabel and shoot that to the hub.

Some metrics on the # initialized steps. the # initialized LLMs, # intiialized pipelines, and their throughput.

N.A.

**Additional context**
https://github.com/argilla-io/argilla/pull/5218

davidberenstein1957:
Closes #810

- [x] add documentation
- [ ] capture additional exceptions ( during `run` method of pipeline for example? @plaguss )


tests/integration/test_telemetry.py
```diff
@@ -0,0 +1,173 @@
...
+    {"prompt": "What's the capital of Spain?"},
+    {"prompt": "Tell me a joke"},
+    {"prompt": "Write a short haiku"},
+    {"prompt": "Translate 'My name is Alvaro' to Spanish"},
+    {"prompt": "What's the capital of Spain?"},
+    {"prompt": "Tell me a joke"},
+    {"prompt": "Write a short haiku"},
+    {"prompt": "Translate 'My name is Alvaro' to Spanish"},
+    {"prompt": "What's the capital of Spain?"},
+]
```
plaguss, line 106:
```suggestion
DATA = [
    {"prompt": "Tell me a joke"},
    {"prompt": "Write a short haiku"},
    {"prompt": "Translate 'My name is Alvaro' to Spanish"},
    {"prompt": "What's the capital of Spain?"}
] * 10
```

tests/integration/test_telemetry.py
```diff
@@ -0,0 +1,173 @@
...
+def test_pipeline_telemetry(mock_telemetry) -> None:
+    with Pipeline(name="unit-test-pipeline") as pipeline:
+        load_dataset = LoadDataFromDicts(name="load_dataset", data=DATA, batch_size=8)
+        rename_columns = RenameColumns(name="rename_columns", input_batch_size=12)
+        generate_response = GenerateResponse(
+            name="generate_response", input_batch_size=16
+        )
+
+        load_dataset.connect(rename_columns)
+        rename_columns.connect(generate_response)
```
plaguss, line 158:
```suggestion
        load_dataset >> rename_columns >> generate_response
```

plaguss (changes requested):
Instead of adding the calls to `add_step` and `add_edge`, we could send the information of `pipeline.dump()` on `BasePipeline.run` so we have all the relevant information we already use for serialization/caching? It would be a single call containing all the information from the pipeline.

I think it could be useful tracking the exceptions raised from the pipeline as you mention, maybe in the `_StepWrapper` would be easier, as that's the container in charge of running the steps on each process.

plaguss:
Not sure how are we going to access the data afterwards to generate the data, I guess it depends, but accessing the dump of the pipeline and cleaning the content we don't want from there I think it's easier, at least from the point of view of `distilabel`, not sure if this is the best way of accessing tracking the telemetry data.

For the exceptions I'm really not sure of what should we track, I guess everything that it's a user error, so you are right not tracking at the `_StepWrapper` level.

davidberenstein1957:
@plaguss, I will do the pipeline thingies based on the dump and make individual calls from the telemetry clients instead.

@gabrielmbmb any specific exceptions you would like to see captured? I capture the RuntimeErrors but perhaps some users errors would be interesting, not sure what we can capture cleanly without filling the code up with try-excepts. 

Also, we can start with this and consider adding more at a later stage, in case we feel miss anything.

davidberenstein1957:
> @plaguss, I will do the pipeline thingies based on the dump and make individual calls from the telemetry clients instead.
> 
> @gabrielmbmb any specific exceptions you would like to see captured? I capture the RuntimeErrors but perhaps some users errors would be interesting, not sure what we can capture cleanly without filling the code up with try-excepts.
> 
> Also, we can start with this and consider adding more at a later stage, in case we feel miss anything.

@plaguss, I updated the code. Decided to leave the batching and we lost some tracking differentiation for generator, normal and global steps.

PR #920, opened 2024-08-21
Add `vLLMEmbeddings`

plaguss:
## Description

Adds a new embeddings provider:

```python
from distilabel.embeddings import vLLMEmbeddings

embeddings = vLLMEmbeddings(model="intfloat/e5-mistral-7b-instruct")

embeddings.load()

results = embeddings.encode(inputs=["distilabel is awesome!", "and Argilla!"])
# [
#   [-0.05447685346007347, -0.01623094454407692, ...],
#   [4.4889533455716446e-05, 0.044016145169734955, ...],
# ]
```

Tested on the cluster and works nicely outside of a `Pipeline`, on a `Pipeline` with `ray` fails, not sure if that's due to `ray` or the `CudaDevicePlacement`. Still need to investigate that, but this should be ready

### DIFF
--- a/src/distilabel/embeddings/__init__.py
+++ b/src/distilabel/embeddings/__init__.py
@@ -14,8 +14,10 @@
 
 from distilabel.embeddings.base import Embeddings
 from distilabel.embeddings.sentence_transformers import SentenceTransformerEmbeddings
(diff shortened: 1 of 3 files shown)


gabrielmbmb (approved):
LGTM!

PR #931, opened 2024-08-27
Add minhash related steps to deduplicate texts

plaguss:
## Description

This PR adds two new steps: `MinHash` and `MinHashLSH` to deduplicate text:

```python
from distilabel.steps import LoadDataFromDicts, MinHash, MinHashLSH


def test_minhash_deduplication():
    with Pipeline() as pipeline:
        data = LoadDataFromDicts(
            data=[
                {"text": "This is a test document."},
                {"text": "This document is a test."},
                {"text": "Test document for duplication."},
                {"text": "Document for duplication test."},
                {"text": "This is another unique document."},
            ]
        )
        minhash = MinHash(tokenizer="ngrams", n=1)
        minhash_lsh = MinHashLSH(threshold=0.9, seed=minhash.seed, drop_hashvalues=True)
        data >> minhash >> minhash_lsh

    ds_dedup = ds.filter(lambda x: x["minhash_duplicate"] is False)
    assert len(ds_dedup) == 4
```

Internally we are using [`datasketch`](https://github.com/ekzhu/datasketch), and to construct the LSH index it relies on a python dict, which limits the size of the datasets we can handle to the RAM. They offer different choices to scale, either `redis` or `cassandra`, for our case we would be better if we implement something that works on disk and is simpler than deploying those solutions, even if it's slower


src/distilabel/steps/filtering/minhash.py
```diff
@@ -0,0 +1,313 @@
...
+
+def tokenize_on_ngrams(texts: Iterable[str], n: int = 1) -> List[Set[bytes]]:
+    """Tokenizes a list of texts into ngrams, and returns the set of them as bytes.
+
+    Args:
+        texts: List of documents to be tokenized.
+        n: The size of the ngrams, defaults to 1 (single letters).
+
+    Returns:
+        Iterable[Set[bytes]]: List with the set of tokens for each document.
```
gabrielmbmb, line 76:
```suggestion
        List with the set of tokens for each document.
```

src/distilabel/steps/filtering/minhash.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+def tokenized_on_words(texts: Iterable[str]) -> List[Set[bytes]]:
+    """Tokenizes a list of texts into words, using `nltk.word_tokenize`.
+
+    Args:
+        texts: List of documents to be tokenized.
+
+    Returns:
+        Iterable[Set[bytes]]: List with the set of tokens for each document.
```
gabrielmbmb, line 63:
```suggestion
        List with the set of tokens for each document.
```

src/distilabel/steps/filtering/minhash.py
```diff
@@ -0,0 +1,313 @@
...
+            "This is another unique document."
+        ]
+        from distilabel.steps import MinHash
+        minhasher = MinHash(tokenizer="ngrams", n=3)
+        minhasher.load()
+        result = next(hasher.process([{"text": t} for t in texts]))
+        ```
+    """
+
+    column: str = "text"
```
gabrielmbmb, line 144: why do we need this attribute? We should handle these cases using `input_mappings`

src/distilabel/steps/filtering/minhash.py
```diff
@@ -0,0 +1,313 @@
...
+
+from distilabel.steps.base import GlobalStep, Step, StepInput
+
+if TYPE_CHECKING:
+    from datasketch import LeanMinHash, MinHash, MinHashLSH
+
+    from distilabel.steps.typing import StepOutput
+
+
+_DUPLICATE_COLUMN: Final[str] = "minhash_duplicate"
```
gabrielmbmb, line 43: I think we should follow a naming convention for the filtering steps. Maybe it's better if we call the column `keep_row_after_<filtering>_filtering`? `True` if the column should be kept, `False` otherwise.

The filtering expression would be also simpler: `ds.filter(lambda x: x["keep_row_after_<filtering>_filtering"])`

tests/integration/test_deduplication.py
```diff
@@ -0,0 +1,51 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from distilabel.pipeline import Pipeline
+from distilabel.steps import LoadDataFromDicts, MinHash, MinHashLSH
+
+
+def test_minhash_deduplication():
```
gabrielmbmb, line 19:
```suggestion
def test_minhash_deduplication() -> None:
```

src/distilabel/steps/filtering/minhash.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+# Copied from: https://github.com/huggingface/datatrove/blob/main/src/datatrove/utils/text.py#L89C1-L95C65
+def ngrams(sequence: Iterable, n: int):
```
gabrielmbmb, line 47:
```suggestion
def ngrams(sequence: Iterable[str], n: int) -> Iterator[Tuple[str, ...]]:
```

src/distilabel/steps/filtering/minhash.py
```diff
@@ -0,0 +1,313 @@
...
+
+import importlib
+import uuid
+from functools import partial
+from itertools import tee
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Final,
+    Iterable,
```
gabrielmbmb, line 23:
```suggestion
    Iterable,
    Iterator,
```

gabrielmbmb (approved):
LGTM! Just some comments regarding the columns and type hints.

gabrielmbmb:
We also need installing the `minhash` extra in the `install_dependencies.sh` script.

src/distilabel/steps/filtering/minhash.py
plaguss: perfect

src/distilabel/steps/filtering/minhash.py
plaguss: right, there's no need, I was using `input_mappings` anyway...

PR #937, opened 2024-08-30
Refactor of MinHash to work with a single class and fix the shelve backend

plaguss:
## Description

This PR does a refactor of the minhash deduplication mechanism:

`MinHash` and `MinHashLSH` now work with a single class `MinHashDedup` which does everything.
When `storage="disk"`, we use `diskcache` instead of `shelve`. I couldn't make it work with `shelve` in macos, due to `dbm` limitations, and in `linux` after installing `gdbm`, the step never advanced at all.

```python
from distilabel.steps import MinHashDedup
from distilabel.steps import LoadDataFromDicts

with Pipeline() as pipeline:
    ds_size = 1000
    batch_size = 500  # Bigger batch sizes work better for this step
    data = LoadDataFromDicts(
        data=[
            {"text": "This is a test document."},
            {"text": "This document is a test."},
            {"text": "Test document for duplication."},
            {"text": "Document for duplication test."},
            {"text": "This is another unique document."},
        ]
        * (ds_size // 5),
        batch_size=batch_size,
    )
    minhash_dedup = MinHashDedup(
        tokenizer="words",
        threshold=0.9,      # lower values will increase the number of duplicates
        storage="dict",     # or "disk" for bigger datasets
    )

    data >> minhash_dedup

if __name__ == "__main__":
    # Filter out the duplicates
    ds_dedup = ds.filter(lambda x: x["keep_row_after_minhash_filtering"])
```


src/distilabel/steps/filtering/minhash.py
```diff
@@ -81,154 +80,56 @@ def tokenize_on_ngrams(texts: Iterable[str], n: int = 1) -> List[Set[bytes]]:
...
-# Also, instead of returning the values, we could be saving them as artifacts,
-# This still needs to be studied.
-class MinHash(Step):
-    """Creates the components for a `MinHash` object to deduplicate texts.
+class MinHashDedup(Step):
+    """Deduplicates text using `MinHash` and `MinHashLSH`.
 
     From `datasketch` documentation:
     Estimates the Jaccard similarity (resemblance) between sets of arbitrary sizes in linear
     time using a small and fixed memory space.
```
gabrielmbmb, line 88: Let's try to use the format of: "`MinHashDedup` is an `Step` that..."

src/distilabel/steps/filtering/minhash.py
```diff
@@ -81,154 +80,56 @@ def tokenize_on_ngrams(texts: Iterable[str], n: int = 1) -> List[Set[bytes]]:
...
         - text (`str`): the texts to be filtered.
-        - hashvalues (`List[int]`): hash values obtained from `MinHash` step.
 
     Output columns:
-        - minhash_duplicate (`bool`): boolean indicating if the piece of text is a
-            duplicate or not, so the user can decide afterwards whether to remove it
-            or not.
+        - keep_row_after_minhash_filtering (`bool`): boolean indicating if the piece of text is a
+            duplicate or not, so the user can decide afterwards whether to remove it.
+            If `True`, the text is not a duplicate.
```
gabrielmbmb, line 115:
```suggestion
        - keep_row_after_minhash_filtering (`bool`): boolean indicating if the piece `text` is
            not a duplicate i.e. this text should be kept.
```

gabrielmbmb (approved):

PR #940, opened 2024-09-02
fix: regex expression in POSITIVE_NEGATIVE

sdiazlor:
Just a very small change to the POSITIVE_NEGATIVE_PAIR_REGEX. Checking the raw output, for instance, here: https://huggingface.co/datasets/distilabel-internal-testing/example-retrieval-reranking-dataset?row=0. Don't know why, but the LLM adds a \s just before ##Positive. So, it was not possible to capture it without this small change.


plaguss:
Maybe is the chosen LLM, that adds that. I tested the task with a stronger model and didn't found the problem. Thanks for the fix!

plaguss (approved):

PR #946, opened 2024-09-04
Add step to deduplicate records based on embeddings

plaguss:
## Description

This PR adds an `EmbeddingDedup` step to deduplicate texts based on embeddings, and previously detected nearest neigbours. The typical workflow is the following:

```python
with Pipeline() as pipeline:
    loader = LoadDataFromDicts(
        data=...,  # Records containing the texts and embeddings
    )

    # NOTE: Guide to choose an index: https://github.com/facebookresearch/faiss/wiki/Guidelines-to-choose-an-index
    nn = FaissNearestNeighbour(
        k=3,
        metric_type=faiss.METRIC_INNER_PRODUCT,
        search_batch_size=50,
        # This fields would be needed in case of training an index (a big dataset)
        # string_factory="IVF300_HNSW32,Flat",
        # train_size=len(dataset),
    )

    embedding_dedup = EmbeddingDedup(
        threshold=0.99,
    )
    loader >> nn >> embedding_dedup

ds_dedup = ds.filter(lambda x: x["keep_row_after_embedding_filtering"])
```

### DIFF
--- a/src/distilabel/steps/filtering/embedding.py
+++ b/src/distilabel/steps/filtering/embedding.py
@@ -171,6 +171,7 @@ class EmbeddingDedup(GlobalStep):
             indices = np.array(input["nn_indices"])[indices_scores]
             if len(indices) > 0:  # If there are any rows found over the threshold
                 rows_to_remove.update(list(indices))
+
         # Remove duplicates and get the list of rows to remove
         for idx in rows_to_remove:
             inputs[idx]["keep_row_after_embedding_filtering"] = False
--- a/src/distilabel/steps/filtering/embedding.py
+++ b/src/distilabel/steps/filtering/embedding.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Optional
 
 import numpy as np
+from pydantic import Field
 from rich.progress import track
 from typing_extensions import override
 
+from distilabel.mixins.runtime_parameters import RuntimeParameter
 from distilabel.steps.base import GlobalStep, StepInput
 
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


src/distilabel/steps/filtering/embedding.py
```diff
@@ -0,0 +1,179 @@
...
+
+        if __name__ == "__main__":
+            distiset = pipeline.run(use_cache=False)
+            ds = distiset["default"]["train"]
+            # Filter out the duplicates
+            ds_dedup = ds.filter(lambda x: x["keep_row_after_embedding_filtering"])
+        ```
+    """
+
+    threshold: float = 0.9
```
gabrielmbmb, line 154: do you think it would be good to have this as a `RuntimeParameter`?

src/distilabel/steps/generators/huggingface.py
```diff
@@ -243,20 +243,19 @@ def _dataset_info(self) -> Dict[str, DatasetInfo]:
...
-        repo_id = self.repo_id
-        config = self.config
 
         try:
-            return get_dataset_infos(repo_id)
+            return get_dataset_infos(self.repo_id)
         except Exception as e:
             # The previous could fail in case of a internet connection issues.
             # Assuming the dataset is already loaded and we can get the info from the loaded dataset, otherwise it will fail anyway.
+            # TODO: This fails when the step is not loaded using make_generator_step.
```
gabrielmbmb, line 252: is this something you have to tackle on this PR?

gabrielmbmb (approved):
LGTM! Much needed step :)

src/distilabel/steps/filtering/embedding.py
plaguss: Yes, will let the option too

src/distilabel/steps/generators/huggingface.py
plaguss: It's done, forgot to remove the comment!

PR #948, opened 2024-09-05
Add `TextClassification`, `UMAP`, `DBSCAN` and `TextClustering` tasks

plaguss:
## Description

This PR adds a general `TextClassification` task, which can be useful for both label and multilabel classification.

Also `UMAP`, `DBSCAN` and `TextClustering`, which are quite related to each other, to generate clusters of text.

It was defined as part of a pipeline for text clustering, so may be biased towards that definition, but should be general enough to work on any common text classification tasks.

**TODO**. Create an example in the docs of a full text clustering pipeline, including the label inference.

### DIFF
--- a/src/distilabel/steps/clustering/dbscan.py
+++ b/src/distilabel/steps/clustering/dbscan.py
@@ -135,6 +135,9 @@ class DBSCAN(GlobalStep):
             n_jobs=self.n_jobs,
         )
 
+    def unload(self) -> None:
+        self._clusterer = None
+
     @property
     def inputs(self) -> List[str]:
         return ["projection"]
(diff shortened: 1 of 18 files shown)


src/distilabel/steps/clustering/dbscan.py
```diff
@@ -0,0 +1,174 @@
...
+    StepInput,
+)
+
+if TYPE_CHECKING:
+    from sklearn.cluster import DBSCAN as _DBSCAN
+
+    from distilabel.steps.typing import StepOutput
+
+
+class DBSCAN(GlobalStep):
```
gabrielmbmb, line 33: Lets add the `unload` method that set the `self._clusterer = None`

src/distilabel/steps/clustering/umap.py
```diff
@@ -0,0 +1,161 @@
...
+    StepInput,
+)
+
+if TYPE_CHECKING:
+    from umap import UMAP as _UMAP
+
+    from distilabel.steps.typing import StepOutput
+
+
+class UMAP(GlobalStep):
```
gabrielmbmb, line 33: Add `unload` method to set `self._umap = None`

gabrielmbmb (changes requested):
LGTM!

PR #973, opened 2024-09-12
[DOCS] Add tips in the docs to avoid overloading Free Serverless Endpoints

plaguss:
## Description

To avoid unexpected errors due to overloading the Free Serverless Endpoints, we can just lower the input batch size, this PR adds hints in the docs to help the users it.

Title entry in the components gallery:

image

FAQ entry:

image

### DIFF
--- a/src/distilabel/llms/huggingface/inference_endpoints.py
+++ b/src/distilabel/llms/huggingface/inference_endpoints.py
@@ -74,7 +74,7 @@ class InferenceEndpointsLLM(AsyncLLM, MagpieChatTemplateMixin):
         `:hugging:`
 
     Examples:
-        Free serverless Inference API (set the `input_batch_size=1` to avoid overloadings!):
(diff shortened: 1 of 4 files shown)


src/distilabel/llms/huggingface/inference_endpoints.py
```diff
@@ -74,13 +74,13 @@ class InferenceEndpointsLLM(AsyncLLM, MagpieChatTemplateMixin):
         `:hugging:`
 
     Examples:
-        Free serverless Inference API:
+        Free serverless Inference API (set the `input_batch_size=1` to avoid overloadings!):
```
dvsrepo, line 77: I'd be more specific:

Set the input_batch_size of the Task that uses this to avoid Model is overloaded.

Important for SEO and search: use the exact wording of the error (i think is as you put it in the FAQ but good to double check)

PR #988, opened 2024-09-18
Updated PushToHub to support merging datasets

dameikle:
Not sure if this is something you'd want upstream but added support for merging datasets to PushToHub


plaguss:
Thanks for the PR @dameikle! But we want to fully rewrite that step to work on batches, but will take into account this behaviour, closing it for the moment

PR #989, opened 2024-09-19
fix missing batch when last batch arrive early

zye1996:
Fix potential missing batch. For instance, batches 1, 2, and 3 arrive before batch 0 and are sent to the next step. In this case, batch 0 can never be sent to the next step as the last batch has been sent.


gabrielmbmb:
Hi @zye1996, good catch! Thank you so much for the PR.

gabrielmbmb (approved):

PR #1000, opened 2024-09-25
Feat/954 llama cpp

bikash119:
This PR adds llama-cpp support to create embeddings.
```
  from distilabel.embeddings import LlamaCppEmbeddings

  embeddings = LlamaCppEmbeddings(model="second-state/all-MiniLM-L6-v2-Q2_K.gguf")

  embeddings.load()

  results = embeddings.encode(inputs=["distilabel is awesome!", "and Argilla!"])
  # [
  #   [-0.05447685346007347, -0.01623094454407692, ...],
  #   [4.4889533455716446e-05, 0.044016145169734955, ...],
  # ]
  ```

### CODE REVIEW

**Comment by bikash119** [comment]
@davanstrien : I have added a working version for llama-cpp support to generate embeddings. Will add more parameters as we have [here](https://github.com/argilla-io/distilabel/blob/main/src/distilabel/llms/llamacpp.py)

**Comment by bikash119** [comment]
- Create embeddings using model from local disk
- Create embeddings using huggingface hub model
- Normalize embeddings by passing `normalize_embeddings=True` to LlamaCppEmbeddings class

**Comment by bikash119** [comment]
@davidberenstein1957 : May I request you to review the changes. In addition, I couldn't add the documentation to `Component Library` or to `API Reference`. I need a little help to add / update documentation.

File: pyproject.toml [hunk 1]
```diff
@@ -94,6 +94,7 @@ vllm = [
     "setuptools",
 ]
 sentence-transformers = ["sentence-transformers >= 3.0.0"]
+llama_cpp_python = ["llama-cpp-python >= 0.2.90"]
```
davidberenstein1957, line 97:
```suggestion
```
we already added this above to the extras

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,179 @@
...
+                    verbose=self.verbose,
+                    embedding=True,
+                    kwargs=self.extra_kwargs,
+                )
+                self._logger.info("Model loaded successfully from Hugging Face Hub")
+            except Exception as e:
+                self._logger.error(
+                    f"Failed to load model from Hugging Face Hub: {str(e)}"
+                )
+                raise
```
davidberenstein1957, line 133: would you be able to capture this logic in the original LLM class too? WDYT @gabrielmbmb

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,179 @@
...
+        else:
+            try:
+                self._logger.info(f"Attempting to load model from: {self.model_name}")
```
davidberenstein1957, line 136: We can skip these additional logs

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,179 @@
...
+
+        results = embeddings.encode(inputs=["distilabel is awesome!", "and Argilla!"])
+        # [
+        #   [-0.05447685346007347, -0.01623094454407692, ...],
+        #   [4.4889533455716446e-05, 0.044016145169734955, ...],
+        # ]
+        ```
+    """
+
+    model: RuntimeParameter[str] = Field(
```
davidberenstein1957, line 67: I don't think this is a RuntimeParameter because the models should always be loaded and not swapped out.

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,179 @@
...
+
+from distilabel.embeddings.base import Embeddings
+from distilabel.llms.mixins.cuda_device_placement import CudaDevicePlacementMixin
+from distilabel.mixins.runtime_parameters import RuntimeParameter
+
+if TYPE_CHECKING:
+    from llama_cpp import Llama as _LlamaCpp
+
+
+class LlamaCppEmbeddings(Embeddings, CudaDevicePlacementMixin):
```
davidberenstein1957, line 27: @bikash119 perhaps you can also take a look at the general structure of the llama cpp llm defintion in terms of argument naming etc :)

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,179 @@
...
+        """Unloads the `gguf` model."""
+        CudaDevicePlacementMixin.unload(self)
+        super().unload()
+
+    @property
+    def model_name(self) -> str:
+        """Returns the name of the model."""
+        return self.model
+
+    def encode(self, inputs: List[str]) -> List[List[Union[int, float]]]:
```
davidberenstein1957, line 159: We don't need the additional ValueErrors and checks here, we can mimick from the logic used in `SentenceTransformerEmbeddings` and `vLLM`.

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,179 @@
...
+            self._logger.error("Model is not initialized")
+            raise ValueError(
+                "Model is not initialized. Please check the initialization process."
+            )
+
+        try:
+            embeds = self._model.embed(inputs, normalize=self.normalize_embeddings)
+            return embeds
+        except Exception as e:
+            print(f"Error creating embedding: {str(e)}")
```
davidberenstein1957, line 178: the try except is not needed and always avoid print statements

tests/unit/conftest.py
```diff
@@ -102,3 +102,14 @@ class DummyTaskOfflineBatchGeneration(DummyTask):
...
+
+@pytest.fixture
+def local_llamacpp_model_path():
+    """
+    Fixture that provides the local model path for LlamaCpp testing.
+
+    Returns:
+        str: The path to the local LlamaCpp model file.
+    """
+    return "./tests/model/gguf/all-MiniLM-L6-v2-Q2_K.gguf"
```
davidberenstein1957, line 115: do you clean this up after the tests?

davidberenstein1957 (commented):
Looking good already. I left some initial comments. Have you seen this already https://github.com/argilla-io/distilabel/pull/999?

tests/unit/conftest.py
bikash119: Thank you @davidberenstein1957  for your comments. I have updated the test. 
It will look for a model in location defined by environment variable `LLAMACPP_TEST_MODEL_PATH`. If the env variable is not set, the test will download a small model from hub , execute the test and then perform cleanup. It uses the `tmp_path` fixture provided by pytest to get a temporary directory for storing the downloaded model.

src/distilabel/embeddings/llamacpp.py
bikash119: @davidberenstein1957 : incorporated the feedback

src/distilabel/embeddings/llamacpp.py
bikash119: @davidberenstein1957 : incorporated the feedback

src/distilabel/embeddings/llamacpp.py
bikash119: @davidberenstein1957 : incorporated the feedback. In addition will keep the a note `avoid print statements` to all my further PRs.

src/distilabel/embeddings/llamacpp.py
bikash119: @davidberenstein1957 : incorporated the feedback

src/distilabel/embeddings/llamacpp.py
bikash119: @davidberenstein1957 : Will wait for comments from @gabrielmbmb . If accepted, will make the changes in another PR. Let me know what you think?

src/distilabel/embeddings/llamacpp.py
bikash119: Thank you @davidberenstein1957  for your review comments. 
- Updated the argument naming and added parameters to keep it consistent with llamacpp definition in llm.
- Updated testcase to test batch embeddings.

src/distilabel/embeddings/llamacpp.py
bikash119: @davidberenstein1957 : One challenge with this approach is, to load model from hub , it use `from_pretrained(...)` api from `llama_cpp`. The `from_pretrained` api does not accepts `token` parameter. So gated model which need hf_token cannot be used for generating embeddings. Let me handle this scenario too or please let me know if you have different thoughts.

bikash119: @davidberenstein1957 : I have updated the code to handle downloading of gated models using HF Token. Please let me know if you find something missing.

bikash119: @davidberenstein1957 : Should I capture this logic in original LLM class too? 
cc : @gabrielmbmb

davidberenstein1957:
Hi Bikash.That would be great. Do you think we can inherent the same / similar logic using a mixin? Perhaps we can tackle a similar approach for vLLM too? Not entirely sure about potential edge cases though but feel free to explore it. Also it might be cleaner to create a separate PR/issue and tackle them there. Regards,DavidOn 1 Oct 2024, at 00:29, bikash119 ***@***.***> wrote:﻿
@bikash119 commented on this pull request.



In src/distilabel/embeddings/llamacpp.py:
> +        if self.hub_repository_id is not None:
+            try:
+                from huggingface_hub.utils import validate_repo_id
+
+                validate_repo_id(self.hub_repository_id)
+            except ImportError as ie:
+                raise ImportError(
+                    "Llama.from_pretrained requires the huggingface-hub package. "
+                    "You can install it with `pip install huggingface-hub`."
+                ) from ie
+            try:
+                self._logger.info(
+                    f"Attempting to load model from Hugging Face Hub: {self.hub_repository_id}"
+                )
+                self._model = _LlamaCpp.from_pretrained(
+                    repo_id=self.hub_repository_id,
+                    filename=self.model,
+                    verbose=self.verbose,
+                    embedding=True,
+                    kwargs=self.extra_kwargs,
+                )
+                self._logger.info("Model loaded successfully from Hugging Face Hub")
+            except Exception as e:
+                self._logger.error(
+                    f"Failed to load model from Hugging Face Hub: {str(e)}"
+                )
+                raise

@davidberenstein1957 : Should I capture this logic in original LLM class too?
cc : @gabrielmbmb

bikash119:
@davidberenstein1957 : Ok , let me create a different PR to handle the scenario. I will keep the this PR untouched and once we have the scenario handled using mixins, will remove the token handling code from `LlamaCppEmbeddings` class.

bikash119:
hi @davidberenstein1957 : I have created a initial version of the mixin to handle model downloads from hub. Please share your feedback.

bikash119:
Hi @davidberenstein1957 : I have made the changes to make model download from hub reusable through mixin. Please share your feedback. If this looks good, will create another PR to handle the model download in llamacpp class of llm too.

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,153 @@
...
+
+        results = embeddings.encode(inputs=["distilabel is awesome!", "and Argilla!"])
+        # [
+        #   [-0.05447685346007347, -0.01623094454407692, ...],
+        #   [4.4889533455716446e-05, 0.044016145169734955, ...],
+        # ]
+        ```
+    """
+
+    model_file: str
```
davidberenstein1957, line 75: would you be able to align this with the LlamaCppLLM class?
Also, could you order the arguments accordingly?

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,153 @@
...
+                n_ctx=self.n_ctx,
+                n_batch=self.n_batch,
+                verbose=self.verbose,
+                embedding=True,
+                **self.extra_kwargs,
+            )
+            self._logger.info("Model loaded successfully")
+        except Exception as e:
+            self._logger.error(f"Failed to load model: {str(e)}")
+            raise
```
davidberenstein1957, line 132: We can remove the logger methods here

src/distilabel/mixins/hub_downloader.py
```diff
@@ -0,0 +1,89 @@
...
+        except ValueError as ve:
+            raise ValueError(f"Invalid repo_id: {self.repo_id}") from ve
+
+        # Determine the download directory
+        download_dir = os.environ.get("DISTILABEL_MODEL_DIR")
+        if download_dir is None:
+            download_dir = tempfile.gettempdir()
+
+        try:
+            model_path = hf_hub_download(
```
davidberenstein1957, line 79: I think we can align with the llama cpp code https://llama-cpp-python.readthedocs.io/en/latest/#pulling-models-from-hugging-face-hub This also allows for pass updating "huggingface cache directory" and that allow s not to write too much underlying code.

davidberenstein1957 (commented):
Don't forget to update the docs and API references :)

src/distilabel/mixins/hub_downloader.py
bikash119: @davidberenstein1957: To download the hub models using llama-cpp, we need to use `Llama.from_pretrained(...)`. This method doesn't accept hf_token as I could understand from [here](https://llama-cpp-python.readthedocs.io/en/latest/api-reference/#llama_cpp.Llama.from_pretrained)
```
@classmethod
def from_pretrained(
    cls,
    repo_id: str,
    filename: Optional[str],
    additional_files: Optional[List] = None,
    local_dir: Optional[Union[str, os.PathLike[str]]] = None,
    local_dir_use_symlinks: Union[bool, Literal["auto"]] = "auto",
    cache_dir: Optional[Union[str, os.PathLike[str]]] = None,
    **kwargs: Any,
) -> "Llama":
    """Create a Llama model from a pretrained model name or path.
    This method requires the huggingface-hub package.
    You can install it with `pip install huggingface-hub`.

    Args:
        repo_id: The model repo id.
        filename: A filename or glob pattern to match the model file in the repo.
        additional_files: A list of filenames or glob patterns to match additional model files in the repo.
        local_dir: The local directory to save the model to.
        local_dir_use_symlinks: Whether to use symlinks when downloading the model.
        **kwargs: Additional keyword arguments to pass to the Llama constructor.

    Returns:
        A Llama model."""
```
 So if we use the `Llama.from_pretrained` users will not be able to download gated model.
Please correct me if you see gap in my understanding.

davidberenstein1957: @bikash119 good catch. I think we could then manually set `os.environ["HF_TOKEN"] = token` and then cal the  `from_pretrained` :)

bikash119:
hi @davidberenstein1957 : Thank you for your time to review my work. I am tried to incorporate your feedback and did a few improvements.
- Added example for user to use both `public` and `private` model. For `private` models, the user has to configure `export HF_TOKEN=hf...` . 
- Added an option to test `--cpu-only` to test embedding generation on cpu.
- Added examples for public , private, cpu based generation.
Please share your feedback.

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,246 @@
...
+        ```
+
+
+    """
+
+    model_path: str
+
+    repo_id: RuntimeParameter[str] = Field(
+        default=None, description="The Hugging Face Hub repository id.", exclude=True
+    )
```
davidberenstein1957, line 147: if we always need to provide a model path, how would we be able to use the repo id?

davidberenstein1957: I think it would be nicer to combine the arguments in some way.

davidberenstein1957: `pretrained_model_name_or_path ` perhaps? similar to the the tranformers naming.

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,246 @@
...
+            raise ImportError(
+                "`llama-cpp-python` package is not installed. Please install it using"
+                " `pip install llama-cpp-python`."
+            ) from ie
+
+        if self.repo_id is not None:
+            try:
+                from huggingface_hub.utils import validate_repo_id
+
+                validate_repo_id(self.repo_id)
```
davidberenstein1957, line 192: I would move this validation outside the ImportError try except.

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,246 @@
...
+            except ImportError as ie:
+                raise ImportError(
+                    "Llama.from_pretrained requires the huggingface-hub package. "
+                    "You can install it with `pip install huggingface-hub`."
+                ) from ie
+            try:
```
davidberenstein1957, line 198: the try except is not needed here, we can just run the code directly and see what it raises

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,246 @@
...
+                    n_ctx=self.n_ctx,
+                    n_batch=self.n_batch,
+                    verbose=self.verbose,
+                    embedding=True,
+                    kwargs=self.extra_kwargs,
+                )
+            except Exception:
+                raise
+        else:
+            try:
```
davidberenstein1957, line 213: same here about the try except :)

tests/unit/embeddings/test_llamacpp.py
```diff
@@ -0,0 +1,203 @@
...
+import numpy as np
+import pytest
+
+from distilabel.embeddings import LlamaCppEmbeddings
+
+"""
+To test with CPU only, run the following command:
+pytest tests/unit/embeddings/test_llamacpp.py --cpu-only
+
+"""
```
davidberenstein1957, line 25: we can leave this out

tests/unit/conftest.py
```diff
@@ -102,3 +105,82 @@ class DummyTaskOfflineBatchGeneration(DummyTask):
...
+    model_path = tmp_path / model_name
+
+    if not model_path.exists():
+        print(f"Downloading test model to {model_path}...")
+        urlretrieve(model_url, model_path)
+        print("Download complete.")
+
+    def cleanup():
+        if model_path.exists():
+            print(f"Cleaning up downloaded model at {model_path}...")
```
davidberenstein1957, line 162: we can remove these print statements

tests/unit/conftest.py
```diff
@@ -102,3 +105,82 @@ class DummyTaskOfflineBatchGeneration(DummyTask):
...
+def pytest_addoption(parser):
+    """
+    Add a command-line option to pytest for CPU-only testing.
+    """
+    parser.addoption(
+        "--cpu-only", action="store", default=False, help="Run tests on CPU only"
+    )
+
+
+@pytest.fixture
```
davidberenstein1957, line 181: not sure if we need GPU testing, so we can just default to only using CPU

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,246 @@
...
+        embeddings.load()
+
+        results = embeddings.encode(inputs=["distilabel is awesome!", "and Argilla!"])
+        # [
+        #   [-0.05447685346007347, -0.01623094454407692, ...],
+        #   [4.4889533455716446e-05, 0.044016145169734955, ...],
+        # ]
+        ```
+
+
```
davidberenstein1957, line 140:
```suggestion
        ```
```

davidberenstein1957 (commented):
Thanks for the updates. Left some minor comments. :)

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,251 @@
...
+from distilabel.embeddings.base import Embeddings
+from distilabel.llms.mixins.cuda_device_placement import CudaDevicePlacementMixin
+from distilabel.mixins.runtime_parameters import RuntimeParameter
+
+if TYPE_CHECKING:
+    from llama_cpp import Llama
+
+
+class LlamaCppEmbeddings(Embeddings, CudaDevicePlacementMixin):
```
davidberenstein1957, line 28: aren't some of the attributes already present in the parent class?

src/distilabel/embeddings/llamacpp.py
```diff
@@ -0,0 +1,251 @@
...
+                    n_gpu_layers=self.n_gpu_layers,
+                    seed=self.seed,
+                    n_ctx=self.n_ctx,
+                    n_batch=self.n_batch,
+                    verbose=self.verbose,
+                    embedding=True,
+                    kwargs=self.extra_kwargs,
+                )
+            except Exception:
+                raise
```
davidberenstein1957, line 228:
```suggestion
            self._model = Llama(
                model_path=str(Path(self.model_path) / self.model),
                n_gpu_layers=self.n_gpu_layers,
                seed=self.seed,
                n_ctx=self.n_ctx,
                n_batch=self.n_batch,
                verbose=self.verbose,
                embedding=True,
                kwargs=self.extra_kwargs,
            )
```

tests/unit/conftest.py
```diff
@@ -102,3 +105,78 @@ class DummyTaskOfflineBatchGeneration(DummyTask):
...
+    Example:
+        export LLAMACPP_TEST_MODEL_PATH="/path/to/your/model.gguf"
+
+    Args:
+        tmp_path_factory: Pytest fixture providing a temporary directory factory.
+
+    Returns:
+        str: The path to the local LlamaCpp model file.
+    """
+    print("\nLlamaCpp model path information:")
```
davidberenstein1957, line 132:
```suggestion
```

tests/unit/conftest.py
```diff
@@ -102,3 +105,78 @@ class DummyTaskOfflineBatchGeneration(DummyTask):
...
+
+    # Check for environment variable first
+    env_path = os.environ.get("LLAMACPP_TEST_MODEL_PATH")
+    if env_path:
+        print(f"Using custom model path from LLAMACPP_TEST_MODEL_PATH: {env_path}")
```
davidberenstein1957, line 137:
```suggestion
```

tests/unit/conftest.py
```diff
@@ -102,3 +105,78 @@ class DummyTaskOfflineBatchGeneration(DummyTask):
...
+        if not os.path.exists(env_path):
+            raise FileNotFoundError(
+                f"Custom model file not found at {env_path}. Please ensure the file exists."
+            )
+        return env_path
+
+    print("LLAMACPP_TEST_MODEL_PATH not set. Using default test model.")
+    print(
+        "To use a custom model, set the LLAMACPP_TEST_MODEL_PATH environment variable to the path of your model file."
+    )
```
davidberenstein1957, line 147:
```suggestion
```

tests/unit/embeddings/test_llamacpp.py
```diff
@@ -0,0 +1,204 @@
...
```
davidberenstein1957, line 25:
```suggestion
```

tests/unit/embeddings/test_llamacpp.py
```diff
@@ -0,0 +1,204 @@
...
+            repo_id=self.repo_id,
+            model=self.model_name,
+            normalize_embeddings=True,
+            n_gpu_layers=n_gpu_layers,
+        )
+        inputs = [
+            "Hello, how are you?",
+            "What a nice day!",
+            "I hear that llamas are very popular now.",
+        ]
```
davidberenstein1957, line 100: perhaps we can re-use the test data as fixture

tests/unit/embeddings/test_llamacpp.py
davidberenstein1957: @bikash119 I think we only want to run tests on CPU without using CPU

tests/unit/conftest.py
davidberenstein1957: I think we can just fix the model path, download a model from a URL and delete it after running the test. We can just use a very small GGUF model

davidberenstein1957 (approved):
Thanks @bikash119!

PR #1034, opened 2024-10-11
Update `LLM.generate` output to include `statistics`

### LINKED ISSUE #738: [FEATURE] Update `LLM.generate` interface to allow returning arbitrary/extra stuff related to the generation
## Description

Update the `LLM.generate` function signature to allow returning more stuff related to the generation apart from the completion given by the LLM. This would be useful to include generation statistics (number of tokens generated, seconds it took, etc), and to also be able to include intermediate generations (`MixtureOfAgents` generations of the proposers LLM).

plaguss:
## Description

This PR updates the output from `llm.generate` to make it more feature rich.

Previously we only returned the generated text:
```python
GenerateOutput = List[Union[str, None]]
```

Now it is updated to allow for statistics related to the generation:

```python
LLMOutput = List[Union[str, None]]

class TokenCount(TypedDict):
    input_tokens: List[int]
    output_tokens: List[int]

LLMStatistics = Union[TokenCount, Dict[str, Any]]
"""Initially the LLMStatistics will contain the token count, but can have more variables.
They can be added once we have them defined for every LLM.
"""

class GenerateOutput(TypedDict):
    generations: LLMOutput
    statistics: LLMStatistics
```

This PR only includes `input_tokens` and `output_tokens` as statistics, but we can add as much as needed in the future.

This information is moved to `distilabel_metadata` in the following way, to avoid collisions between statistics of different steps:

```python
{
    "generations": ["Hello Magpie"],
    f"statistics_{step_name}": {
        "input_tokens": [12],
        "output_tokens": [12],
    },
}
```

**NOTE**:
Most `Task` reuse the same `Task.process` method to process the generations, and nothing else has to be done, but for tasks like `Magpie` where the `process` method is overwritten, this has to be updated.

Closes #738


src/distilabel/models/llms/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    Args:
+        text_or_messages: Either a string response or a list of messages.
+        tokenizer: A callable function that take str and returns the tokenized version of the text.
+
+    Returns:
+        The number of tokens.
+    """
+    if isinstance(text_or_messages, list):
+        # If it's a list of messages, concatenate the content of each message
+        text = " ".join([message["content"] for message in text_or_messages])
```
gabrielmbmb, line 37: I think it would be better to tokenize each message individually and then sum the results to be 100% precise.

src/distilabel/models/llms/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    generations: "LLMOutput",
+    input_tokens: Optional[List[int]] = None,
+    output_tokens: Optional[List[int]] = None,
+) -> "GenerateOutput":
+    """Helper function to prepare the output of the LLM.
+
+    Args:
+        generations: The outputs from an LLM.
+        input_tokens: The number of tokens of the inputs. Defaults to [0].
+        output_tokens: The number of tokens of the LLM response. Defaults to [0].
```
gabrielmbmb, line 54:
```suggestion
        input_tokens: The number of tokens of the inputs. Defaults to `None`.
        output_tokens: The number of tokens of the LLM response. Defaults to `None`.
```

src/distilabel/models/llms/vllm.py
```diff
@@ -421,6 +446,65 @@ def _prepare_structured_output(
...
+    ) -> "LLMStatistics":
+        output_tokens = [len(output.token_ids) for output in outputs.outputs]
+        return {
+            "input_tokens": [compute_tokens(input, self._tokenizer.encode)]
+            * len(output_tokens),
+            "output_tokens": output_tokens,
+        }
+
+    @staticmethod
+    def _prepare_sorted_resuts(
```
gabrielmbmb, line 460:
```suggestion
    def _prepare_sorted_results(
```

src/distilabel/models/llms/vllm.py
```diff
@@ -383,23 +394,37 @@ def generate(  # type: ignore
...
+                    )
+                )
 
         # If logits_processor is set, we need to sort the outputs back to the original order
         # (would be needed only if we have multiple structured outputs in the dataset)
         if sorted_indices is not None:
-            batched_outputs = _sort_batches(
-                batched_outputs, sorted_indices, num_generations=num_generations
+            # Sort the batched outputs together with the statistics
+            generations = self._prepare_sorted_resuts(
```
gabrielmbmb, line 421:
```suggestion
            generations = self._prepare_sorted_results(
```

src/distilabel/models/llms/vllm.py
```diff
@@ -421,6 +446,65 @@ def _prepare_structured_output(
...
+        statistics = {}
+        for field in statistic_fields:
+            batched_field = _sort_batches(
+                [g["statistics"][field] for g in generations],
+                sorted_indices,
+                num_generations=num_generations,
+            )
+            statistics[field] = batched_field
+
+        # Regenerates the outputs as they are returned buy `preare_output`
```
gabrielmbmb, line 496:
```suggestion
        # Regenerates the outputs as they are returned buy `prepare_output`
```

src/distilabel/steps/clustering/text_clustering.py
```diff
@@ -312,6 +312,7 @@ def process(self, inputs: StepInput) -> "StepOutput":
             self._logger.info(f"📦 Processing internal batch of inputs {i}...")
             results = super().process(batched_inputs)
             for result in next(results):  # Extract the elements from the generator
+                print("INTERMEDIATE RESULTS", result)
```
gabrielmbmb, line 315:
```suggestion
```

src/distilabel/steps/tasks/base.py
```diff
@@ -453,3 +474,64 @@ class GlobalTask(_Task, GlobalStep):
...
+            remainder = gen_length % current_length
+
+            # Create new list with repeated values
+            new_values = stat_values * repeats + stat_values[:remainder]
+            output["statistics"][stat_key] = new_values
+
+    return output
+
+
+def iterate_generations_with_stats(output: "GenerateOutput") -> "GenerateOutput":
```
gabrielmbmb, line 523: I think the return type hint should be `Generator[Tuple[...], None, None]`

src/distilabel/steps/tasks/magpie/base.py
```diff
@@ -245,29 +258,32 @@ def _prepare_conversation_outputs(
...
+    ) -> Tuple[List["ChatType"], List[int], "LLMStatistics"]:
         # Generate an output for the conversations that are still active (no previous `None`s)
         outputs = self.llm.generate(
             inputs=[conversations[idx] for idx in active_indices],
             num_generations=1,
             **self.llm.generation_kwargs,  # type: ignore
         )
+        # Extract the single message from the conversation
+        messages = [output["generations"][0] for output in outputs]
+        statistics = [output["statistics"] for output in outputs]
```
gabrielmbmb, line 270: maybe we can combine this in the same for loop

src/distilabel/steps/typing.py
```diff
@@ -15,7 +15,25 @@
...
+#             {"col1": "val1", "col2": "val2"},
+#             {"col1": "val1", "col2": "val2"},
+# 		],
+# 		"statistics": {
+#             "llm": {},
+# 	        "time": 12341234,
+#             ...
+# 		}
+# }
+# """
```
gabrielmbmb, line 36: This should be uncommented, right?

src/distilabel/utils/dicts.py
```diff
@@ -33,8 +35,66 @@ def group_dicts(*dicts: Dict[_K, Any]) -> Dict[_K, List[Any]]:
...
+            k: list(chain.from_iterable(v)) for k, v in combined_dict.items()
+        }
+    return combined_dict
 
 
 def flatten_dict(x: Dict[Any, Any]) -> Dict[Any, Any]:
     return {k: json.dumps(v) if isinstance(v, dict) else v for k, v in x.items()}
+
+
+def merge_dicts(*dict_lists):
```
gabrielmbmb, line 51:
```suggestion
def merge_dicts(*dict_lists: dict) -> list[dict]:
```

gabrielmbmb (approved):

PR #1050, opened 2024-11-01
docs: improve docs for `MinHashDedup` `Step`

anakin87:
:wave: 

While looking at this nice feature, I noticed some errors in the docstring and I'm trying to fix them.

_Slightly related feedback_
Initially I was bit surprised that I cannot easily understand why a single text element is considered a duplicate when using this technique, while this is transparent using `EmbeddingGeneration` + `EmbeddingDedup`.
But I understand that this technique is more complex, even from the implementation point of view...

@gabrielmbmb @plaguss


plaguss (approved):
Thanks a lot for the fixes!

PR #1051, opened 2024-11-04
Fix a tiny typo in `_Step` docstring

sadra-barikbin:
Hi there!

To fix a teeny-tiny typo in `_Step` docstring.


plaguss (approved):
Thanks for the catch!

PR #1060, opened 2024-11-14
Image Language Models and `ImageGeneration` task

plaguss:
## Description

This PR adds a new module to models: `models/image_generation` to store image models (`InferenceEndpointsImageGeneration` and `OpenAIImageGeneration`), with 2 new base classes: `ImageGenerationModel` and `AsyncImageGenerationModel`,  and a new `ImageGeneration` task.

Sample pipeline and [dataset](https://huggingface.co/datasets/plaguss/test-finepersonas-v0.1-tiny-flux-schnell). Take into account the `distiset.transform_columns_to_image` method, necessary to push the dataset with the images as objects instead of strings.

```python
from datasets import load_dataset

from distilabel.models.image_generation import InferenceEndpointsImageGeneration
from distilabel.steps import KeepColumns
from distilabel.steps.tasks import ImageGeneration

ds = load_dataset("dvilasuero/finepersonas-v0.1-tiny", split="train").select(range(3))

with Pipeline(name="image_generation_pipeline") as pipeline:
    igm = InferenceEndpointsImageGeneration(model_id="black-forest-labs/FLUX.1-schnell")

    img_generation = ImageGeneration(
        name="flux_schnell", image_generation_model=igm, input_mappings={"prompt": "persona"}
    )

    keep_columns = KeepColumns(columns=["persona", "model_name", "image"])

    img_generation >> keep_columns


if __name__ == "__main__":
    distiset = pipeline.run(use_cache=False, dataset=ds)
    # Save the images as `PIL.Image.Image`
    distiset = distiset.transform_columns_to_image("image")
    distiset.push_to_hub("plaguss/test-finepersonas-v0.1-tiny-flux-schnell")
```


davidberenstein1957:
@burtenshaw

pyproject.toml
```diff
@@ -102,6 +102,7 @@ text-clustering = [
     "scikit-learn >= 1.4.1",
     "matplotlib >= 3.8.3",   # For the figure (even though it's optional)
 ]
+vision = ["Pillow >= 10.3.0"]  # To work with images.
```
davidberenstein1957, line 105: I just go a pil error due to some imports

```
/Users/davidberenstein/Documents/programming/argilla/distilabel/.venv/lib/python3.11/site-packag │
│ es/distilabel/steps/tasks/text_generation_with_image.py:18 in <module>                           │
│                                                                                                  │
│    15 from typing import TYPE_CHECKING, Any, Literal, Union                                      │
│    16                                                                                            │
│    17 from jinja2 import Template                                                                │
│ ❱  18 from PIL import Image                                                                      │
│    19 from pydantic import Field                                                                 │
│    20                                                                                            │
│    21 from distilabel.steps.tasks.base import Task                                               │
│                                                                                                  │
│ ╭──────────── locals ────────────╮                                                               │
│ │       Literal = typing.Literal │                                                               │
│ │ TYPE_CHECKING = False          │                                                               │
│ │         Union = typing.Union   │                                                               │
│ ╰────────────────────────────────╯                                                               │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ModuleNotFoundError: No module named 'PIL'
```

src/distilabel/models/image_generation/base.py
```diff
@@ -0,0 +1,332 @@
...
+                return runtime_parameters_info
+            for key, value in generation_kwargs_info["optional"].items():
+                info = {"name": key, "optional": value}
+                if description := generate_docstring_args.get(key):
+                    info["description"] = description
+                generation_kwargs_info["keys"].append(info)
+
+            generation_kwargs_info.pop("optional")
+
+        return runtime_parameters_info
```
gabrielmbmb: The issue here is that when calling `super().get_runtime_parameters_info()`, the `InferenceEndpointsLLM.get_runtime_parameters_info` method is being called instead of the one from `RuntimeParametersMixin` class, which is returning the dictionary without the `optional` key (already popped). I guess similar it's happening with `OpenAI` class.

gabrielmbmb: I think the inheritance from `InferenceEndpointsLLM` is a bit messy, not only in this method but also in other parts of the class. For example:

```python
from distilabel.models import InferenceEndpointsImageGeneration, InferenceEndpointsLLM

igm = InferenceEndpointsImageGeneration()
```

This should raise a `ValidationError` because of the validator `only_one_of_model_id_endpoint_name_or_base_url_provided` (no endpoint name or model id), but instead gives `TypeError`: 

```
TypeError: ValidationError.__new__() missing 1 required positional argument: 
'line_errors'
```

which is caused by the multiple inheritance I think.

gabrielmbmb: I think we should:

1. Move extended methods from `RuntimeParametersMixins` like `runtime_parameters_names`, `get_runtime_parameters_info` and `generate_parsed_docstring` to another class that then we can use in the base `LLM`, `ImageGenerationModel` and the future base classes to come.
2. Just duplicate the code for classes that uses a client like inference endpoints or openai, because inheriting from the `LLM` class it's messy or can get messy. Maybe we can create a base class that offers the client functionality for openai, inference endpoints, etc and then use it in the `LLM`s, `ImageGenerationModel`, etc.

gabrielmbmb (commented):
Very cool! I think we need to fix some issues related to inheritance, but maybe we can tackle those in a separate PRs before the release.

src/distilabel/models/image_generation/base.py
```diff
@@ -0,0 +1,247 @@
...
+        synchronously awaiting for the response of each input sent to `agenerate`.
+
+        Args:
+            inputs: the list of inputs to generate responses for.
+            num_generations: the number of generations to generate per input.
+            **kwargs: the additional kwargs to be used for the generation.
+
+        Returns:
+            A list containing the generations for each input.
+        """
```
gabrielmbmb, line 234:
```suggestion
        """Method to generate a list of images asynchronously, returning the output
        synchronously awaiting for the image of each input sent to `agenerate`.

        Args:
            inputs: the list of inputs to generate images for.
            num_generations: the number of generations to generate per input.
            **kwargs: the additional kwargs to be used for the generation.

        Returns:
            A list containing the images for each input.
        """
```

gabrielmbmb (approved):
LGTM!

PR #1066, opened 2024-11-20
Add `TextGenerationWithImage` task

plaguss:
## Description

This PR adds a `VisionGeneration` task to work with image-tex-to-text tasks (chatting with images).

```python
with Pipeline(name="vision_generation_pipeline") as pipeline:
    loader = LoadDataFromDicts(
        data=[
            {
                "instruction": "What’s in this image?",
                "image": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
            }
        ],
    )

    vision = TextGenerationWithImage(
        name="vision_gen",
        llm=InferenceEndpointsLLM(
            model_id="meta-llama/Llama-3.2-11B-Vision-Instruct",
        ),
        image_type="url"
    )

    loader >> vision
```

It has been tested with `InferenceEndpointsLLM` and `OpenAILLM`. There are differences across providers, so we need more tests to ensure the remaining are working, specifically, `vLLM` internally calls to `llm.chat` instead of `llm.generate`, so we need to update it for the vision case


burtenshaw:
@plaguss Just an API level observation, but does it make sense for `VisionGeneration` to be task? because it means that it cannot be combined with other tasks. Like for example, you could not do `SelfInstruct` on an image of a document.

Also `VisionGeneration` needs logic and behaviour that is generally in the `llms` module, like processing messages.

burtenshaw:
Here are some use cases that would require vision language models.
- caption generation
- video caption generation
- DocVQA on pdf (SelfInstruct)
- Judge (ultrafeedback ++) on image generation

src/distilabel/steps/tasks/typing.py
```diff
@@ -15,18 +15,57 @@
...
 
 from pydantic import BaseModel
-from typing_extensions import TypedDict
+from typing_extensions import Required, TypedDict
 
 
 class ChatItem(TypedDict):
-    role: str
+    role: Literal["system", "user", "assistant"]
     content: str
```
gabrielmbmb, line 23: Let's update the type of `content` to something like `Union[str, TextContent, ImageContent]` instead of creating more types like `ConversationType`

src/distilabel/steps/tasks/vision_generation.py
```diff
@@ -0,0 +1,222 @@
...
+        self._template = Template(self.template)
+
+    def _transform_image(self, image: Union[str, "Image"]) -> str:
+        """Transforms the image based on the `image_type` attribute."""
+        if self.image_type == "url":
+            return image
+        elif self.image_type == "base64":
+            return f"data:image/jpeg;base64,{image}"
+        # Othwerwise, it's a PIL image
+        return f"data:image/jpeg;base64,{image_to_str(image)}"
```
gabrielmbmb, line 182:
```suggestion
        if self.image_type == "url":
            return image
            
        if self.image_type == "base64":
            return f"data:image/jpeg;base64,{image}"
            
        # Othwerwise, it's a PIL image
        return f"data:image/jpeg;base64,{image_to_str(image)}"
```

src/distilabel/steps/tasks/vision_generation.py
```diff
@@ -0,0 +1,222 @@
...
+
+        return messages  # type: ignore
+
+
+# TODO: Once we merge the image generation, this function can be reused
+def image_to_str(image: Image.Image, image_format: str = "JPEG") -> str:
+    """Converts a PIL Image to a base64 encoded string."""
+    buffered = io.BytesIO()
+    image.save(buffered, format=image_format)
+    return base64.b64encode(buffered.getvalue()).decode("utf-8")
```
gabrielmbmb, line 222: maybe we can move this to a module `utils/image.py`?

src/distilabel/steps/tasks/vision_generation.py
```diff
@@ -0,0 +1,222 @@
...
+        - dynamic (determined by `columns` attribute): By default will be set to `instruction`.
+            The columns can point both to a `str` or a `list[str]` to be used in the template.
+        - image: The column containing the image URL, base64 encoded image or PIL image.
+
+    Output columns:
+        - generation (`str`): The generated text.
+        - model_name (`str`): The name of the model used to generate the text.
+
+    Categories:
+        - vision-generation
```
gabrielmbmb, line 68: I think we can just simply use `text-generation`

gabrielmbmb (changes requested):
Looks good! Just a few changes needed.

I've discussed with @plaguss, at least for now, to keep the specific task to handle the text generation with an VLM until we found a better interface that allows to use any VLM with any task (such as SelfInstruct).

PR #1086, opened 2024-12-23
Feat/add magpie support llama cpp ollama

davidberenstein1957:
I added extended support for Ollama and Llamacpp.

- ollama support
- llamacpp support
- minor refactors w.r.t. import of `disitlabel.models` module.

I looked into adding OpenAI API format for other providers but this does not work because tokenization is handled server-side and cannot be disabled out of the box.

- litellm - https://docs.litellm.ai/docs/text_completion
- vllm server -https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html?ref=blog.mozilla.ai#id4
- openai - completion is deprecated https://platform.openai.com/docs/api-reference/completions

Perhaps we can refactor the HF InferenceClient a bit to make this work.

# Ollama

```python
from distilabel.models import LlamaCppLLM
from distilabel.steps.tasks import Magpie

llm = LlamaCppLLM(
    model_path="smollm2-360m-instruct-q8_0.gguf",
    tokenizer_id="HuggingFaceTB/SmolLM2-360M-Instruct",
    magpie_pre_query_template="qwen2",
)
magpie = Magpie(
    llm=llm,
)
magpie.load()

print(next(magpie.process(inputs=[{"system": "You are a helpful assistant."}])))
```

# Llamacpp

```python
from distilabel.models import OllamaLLM
from distilabel.steps.tasks import Magpie

llm = OllamaLLM(
    model="llama3.1",
    tokenizer_id="meta-llama/Meta-Llama-3-8B-Instruct",
    magpie_pre_query_template="llama3",
)
magpie = Magpie(llm=llm)
magpie.load()

print(next(magpie.process(inputs=[{"system_prompt": "You are a helpful assistant."}])))
```


src/distilabel/models/llms/ollama.py
```diff
@@ -112,10 +127,35 @@ class OllamaLLM(AsyncLLM):
             description="The structured output format to use across all the generations.",
         )
     )
-
+    tokenizer_id: Optional[RuntimeParameter[str]] = Field(
+        default=None,
+        description="The tokenizer Hugging Face Hub repo id or a path to a directory containing"
```
burtenshaw, line 132:
```suggestion
        description="The Hugging Face Hub repo id or a path to a directory containing"
```

src/distilabel/models/llms/ollama.py
```diff
@@ -112,10 +127,35 @@ class OllamaLLM(AsyncLLM):
...
+        " the tokenizer config files. If not provided, the one associated to the `model`"
+        " will be used.",
+    )
+    use_magpie_template: RuntimeParameter[bool] = Field(
+        default=False,
+        description="Whether to use the Magpie pre-query template or not.",
```
burtenshaw, line 138: Should we explain 'pre-query'?

burtenshaw (approved):
Just some minor comments on documentation, but looks ok.

src/distilabel/models/llms/ollama.py
davidberenstein1957: I think we could but it is set by the magpie task so not relevant for the model and inherited from the Mixin. I removed it to avoid redefinition.

PR #1087, opened 2024-12-24
add approach to flexibly inject system_messages

davidberenstein1957:
I saw the `use_system_prompt` argument was supposed to be removed in the 1.5 release and it seemed like a good excuse to allow from flexible system prompt injection across all tasks. 

I am not sure if it is relevant for things like URIAL, and some system prompts seem to rely on `.format`, so we could disable it there too, but let's iterate a bit on this PR.


davidberenstein1957:
cc @davanstrien  I believe you also vouched for something like this a while ago.

PR #1088, opened 2024-12-26
Support custom default headers in `OpenAILLM` class.

khulaifi95:
Hi this is my first PR on github. I added header customisation to the `OpenAILLM` generic inference class for interacting with OpenAI compatible servers.

This would be useful if the server is privately hosted behind a gateway or proxy, where metadata such as model provider or cloud provider is required to authenticate the request.

The parameter is passed directly to both internal OpenAI clients. The simple test case is added.


davidberenstein1957 (approved):

davidberenstein1957:
@khulaifi95 could you resolve the merge conflicts? After we can merge :)

khulaifi95:
@gabrielmbmb Hi Gabriel, thanks for helping the merge. I figured that conflict was because my commits were on the forked main branch, but the target was set to be the develop branch. Thank you for solving this :).

Issue #995, opened 2024-09-23
[FEATURE] `mlx-lm` integration


gabrielmbmb:
## Description

As suggested by @alvarobartt, it would be nice to integrate [`mlx-lm`](https://pypi.org/project/mlx-lm/)

alvarobartt:
Maybe this is a useful reference https://github.com/dottxt-ai/outlines/blob/515c1979b8507c6d70b07f3b8817e7c24cabfe09/outlines/models/mlxlm.py cc @gabrielmbmb @plaguss

plaguss:
Done in #1089

PR #1092, opened 2025-01-09
update to outlines010

davidberenstein1957:
- [] llama-cpp
- [] transformers waiting on https://github.com/dottxt-ai/outlines/issues/1367
- [] vllm
- [] mlx waiting on https://github.com/argilla-io/distilabel/issues/995

### DIFF
--- a/src/distilabel/steps/tasks/structured_outputs/outlines.py
+++ b/src/distilabel/steps/tasks/structured_outputs/outlines.py
@@ -27,6 +27,7 @@ from typing import (
     get_args,
 )
 
+import pkg_resources
 from pydantic import BaseModel
 
 from distilabel.errors import DistilabelUserError
@@ -36,7 +37,11 @@ if TYPE_CHECKING:
     from distilabel.steps.tasks.typing import OutlinesStructuredOutputType
 
 Frameworks = Literal["transformers", "llamacpp", "vllm"]
-"""Available frameworks for the structured output configuration. """
+# Available frameworks for the structured output configuration.
+_outlines_version = pkg_resources.get_distribution("outlines").version
+outlines_below_0_1_0 = pkg_resources.parse_version(
+    _outlines_version
+) < pkg_resources.parse_version("0.1.0")
 
 
 def model_to_schema(schema: Type[BaseModel]) -> Dict[str, Any]:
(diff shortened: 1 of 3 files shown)


sdiazlor (commented):
Looks good as it's the approach I had started for LlamaCpp.
- Did you check if it worked for LlamaCpp? I had tested before vacations and needed to update `logits_processor=LogitsProcessorList([self._logits_processor]) if self.structured_output else None` in llamacpp.py
- This is the issue I had encountered for Llama models. I guess it should be solved with the previous PR, right? https://github.com/dottxt-ai/outlines/issues/1261

src/distilabel/steps/tasks/structured_outputs/outlines.py
```diff
@@ -46,31 +53,56 @@ def model_to_schema(schema: Type[BaseModel]) -> Dict[str, Any]:
 
 def _get_logits_processor(framework: Frameworks) -> Tuple[Callable, Callable]:
     """Helper function to return the appropriate logits processor for the given framework."""
-    if framework == "transformers":
-        from outlines.integrations.transformers import (
-            JSONPrefixAllowedTokens,
-            RegexPrefixAllowedTokens,
+    if framework not in Frameworks.__args__:
```
burtenshaw, line 56: I find this block confusing and fancy. I would just use an elif below for outlines, then an else to raise the error. 

imo, it'll make the code easier to maintain.

src/distilabel/steps/tasks/structured_outputs/outlines.py
```diff
@@ -46,31 +53,56 @@ def model_to_schema(schema: Type[BaseModel]) -> Dict[str, Any]:
...
-    raise DistilabelUserError(
-        f"Invalid framework '{framework}'. Must be one of {get_args(Frameworks)}",
-        page="sections/how_to_guides/advanced/structured_generation/",
-    )
+
+def _get_outlines_tokenizer_or_model(llm: Any, framework: Frameworks) -> Callable:
+    if outlines_below_0_1_0:
+        return llm
+    else:
+        if framework == "llamacpp":
```
burtenshaw, line 96: Relating to above, you might want to put framework in an `Enum`?

tests/unit/pipeline/.DS_Store
plaguss, line 1: This should be removed

tests/unit/.DS_Store
plaguss, line 1: This should be removed

vllm
plaguss, line 1: what is this?

src/distilabel/steps/tasks/structured_outputs/outlines.py
```diff
@@ -36,7 +37,13 @@
     from distilabel.steps.tasks.typing import OutlinesStructuredOutputType
 
 Frameworks = Literal["transformers", "llamacpp", "vllm"]
-"""Available frameworks for the structured output configuration. """
+
+if importlib.util.find_spec("outlines"):
+    outlines_below_0_1_0 = pkg_resources.parse_version(
+        pkg_resources.get_distribution("outlines").version
+    ) < pkg_resources.parse_version("0.1.0")
```
burtenshaw, line 44: I would move all the code that handles the outlines version/ availability into a single function and have it on these lines, then use it within `prepare_guided_output` . I would also add some logging to that function.

src/distilabel/steps/tasks/structured_outputs/outlines.py
```diff
@@ -97,13 +129,16 @@ def prepare_guided_output(
         case of "json" will also include the schema as a dict, to simplify serialization
         and deserialization.
     """
+
     if not importlib.util.find_spec("outlines"):
```
burtenshaw, line 133: As above, I would move this to a function like `get_outlines_version`.

burtenshaw (commented):
This looks good functionally, but I found it difficult to follow. With community maintainability in mind, I think you could localise the logic about outlines versions.

src/distilabel/steps/tasks/structured_outputs/outlines.py
```diff
@@ -46,31 +53,56 @@ def model_to_schema(schema: Type[BaseModel]) -> Dict[str, Any]:
...
+        from outlines.processors import JSONLogitsProcessor, RegexLogitsProcessor
 
         return JSONLogitsProcessor, RegexLogitsProcessor
 
-    raise DistilabelUserError(
-        f"Invalid framework '{framework}'. Must be one of {get_args(Frameworks)}",
-        page="sections/how_to_guides/advanced/structured_generation/",
-    )
+
+def _get_outlines_tokenizer_or_model(llm: Any, framework: Frameworks) -> Callable:
```
plaguss, line 92: Can we add here a more specific type for the `llm` argument?

plaguss (changes requested):
Looks good! I left some comments. Have you generated some dataset with the 3 integrations to check it works?

burtenshaw (approved):
LGTM

PR #1096, opened 2025-01-13
update: search by match

sdiazlor:


davidberenstein1957:
Hi @sdiazlor, looks good. The user also proposed that this could be needed in other tasks too. Perhaps you can quickly evaluate that.

davidberenstein1957 (approved):

PR #1099, opened 2025-01-16
Update `LLM`s to support prompt logprobs use-case

gabrielmbmb:
## Description

This PR updates `OpenAILLM`, `InferenceEndpointsLLM` and `vLLM` so they return the logprobs of the input prompt. This is useful for the use-case in which an already formatted message (instruction-response) is passed to an LLM to get the prompt logprobs to be later used for training a student model using KD from a teacher model.


Atry:
This PR looks very nice!

However, it seems that logprobs do not appear in TextGeneration or other tasks. Why?

PR #1103, opened 2025-01-17
Fix `MlxLLM` by aligning it with `mlx-lm>=0.21`

davidberenstein1957:


gabrielmbmb (approved):

gabrielmbmb:
Pushed some fixes related class attributes and added some missing docs in the docstring that were disallowing docs to build. Should be good to go.

src/distilabel/models/llms/mlx.py
```diff
@@ -42,7 +44,7 @@ class MlxLLM(LLM, MagpieChatTemplateMixin):
     Attributes:
         path_or_hf_repo: the path to the model or the Hugging Face Hub repo id.
         tokenizer_config: the tokenizer configuration.
-        model_config: the model configuration.
+        mlx_model_config: the MLX model configuration.
```
gabrielmbmb, line 47: We couldn't use `model_config` as it's a reserved attribute of `pydantic.BaseModel` class.

PR #1113, opened 2025-01-27
Fix pipeline getting stuck when multiple step replicas

gabrielmbmb:
## Description

This PR fixes two issues that:

1. The pipeline execution was getting stuck just after receiving the last batch (`batch.last_batch == True`) from the step because we were sending as many `None`s (as flag) as replicas in the step input queue, but some replicas were consuming more than one of these `None`s leaving some replicas waiting forever to get something from the queue.
2. We were finishing the pipeline execution just after receiving the last batch. This is problematic when setting a high number of replicas for one step as it's very likely to receive the batch with `last_batch==True` from one replica before receiving previous batches that could be still being processed by other replicas. To fix this issue, we first check if there's any missing sequence number if step last batch was received.

### DIFF
--- a/src/distilabel/pipeline/batch_manager.py
+++ b/src/distilabel/pipeline/batch_manager.py
@@ -728,6 +728,7 @@ class _BatchManager(_Serializable):
         last_batch_received: Dict[str, Union[_Batch, None]],
         last_batch_sent: Dict[str, Union[_Batch, None]],
         last_batch_flag_sent_to: List[str],
+        received_batch_seq_nos: Dict[str, List[int]],
     ) -> None:
         """Initialize the `_BatchManager` instance.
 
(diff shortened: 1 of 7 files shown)


src/distilabel/pipeline/batch_manager.py
```diff
@@ -778,9 +801,13 @@ def register_batch(
             steps_data_path: The path where the outputs of each `Step` (considering its
                 signature) will be saved for later reuse in another pipelines executions.
         """
-        last_batch = self._last_batch_received[batch.step_name]
+        step_name = batch.step_name
+        seq_no = batch.seq_no
+        self._received_batch_seq_nos[batch.step_name].append(seq_no)
+
+        last_batch = self._last_batch_received[step_name]
```
plaguss, line 808:
```suggestion
        step_name = batch.step_name
        seq_no = batch.seq_no
        self._received_batch_seq_nos[step_name].append(seq_no)

        last_batch = self._last_batch_received[step_name]
```

src/distilabel/pipeline/batch_manager.py
```diff
@@ -778,9 +801,13 @@ def register_batch(
...
         if not last_batch or (last_batch and last_batch.seq_no < batch.seq_no):
```
plaguss, line 809:
```suggestion
        if not last_batch or (last_batch and last_batch.seq_no < seq_no):
```

plaguss (approved):
LGTM. Feel free to skip the corrections

PR #1115, opened 2025-01-30
Corrected AzureOpenAILLM path to mock the internal class

dameikle:
Updated the internal path for mocking the _prepare_structured_output call in the load method


davidberenstein1957 (approved):
@dameikle. thanks for this PR.

kipkoei (approved):
Encountered the same issue, this solves it.

PR #1121, opened 2025-02-15
Fix vLLM v1 compatibility

Atry:
Don't delete model_executor if it does not exist, e.g. when VLLM_USE_V1 is set.

See https://docs.vllm.ai/en/stable/serving/env_vars.html and https://github.com/vllm-project/vllm/blob/main/vllm/v1/engine/llm_engine.py


plaguss (approved):

plaguss:
Thanks for the fix!

PR #1153, opened 2025-05-31
Exclude token from serialization

dameikle:
Based on #1129 from @sysradium this PR excludes all token field from serialization and adds a test

### DIFF
--- a/src/distilabel/utils/serialization.py
+++ b/src/distilabel/utils/serialization.py
@@ -194,7 +194,7 @@ class _Serializable:
         """
         # Any parameter named api_key will be excluded from the dump (those are supposed to be SecretStr anyway,
         # and will remove them afterwards)
-        dump = obj.model_dump(exclude="api_key", **kwargs)
+        dump = obj.model_dump(exclude=("api_key", "token"), **kwargs)
 
         # Check if any attribute in value within the `dump` is an `EnumType`,
         # as it needs a specific serialization.
--- a/src/distilabel/utils/serialization.py
+++ b/src/distilabel/utils/serialization.py
@@ -192,8 +192,8 @@ class _Serializable:
         Returns:
             A dictionary containing the serializable content of the class.
         """
-        # Any parameter named api_key will be excluded from the dump (those are supposed to be SecretStr anyway,
-        # and will remove them afterwards)
+        # Any parameter named api_key or token will be excluded from the dump
+        # (those are supposed to be SecretStr anyway, and will remove them afterwards)
         dump = obj.model_dump(exclude=("api_key", "token"), **kwargs)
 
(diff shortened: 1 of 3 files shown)


sysradium:
Cool. Hopefully that one gets merged :)

dameikle:
Hey @sysradium :wave: - Was getting ready to add a comment asking if you could check this was working for you? I've cherry-picked the commits to keep your contribution and authorship. Really good spot noticing this bug.

sysradium:
Just tested - all good, works for me.
Essentially here is one of the repos where I used it: https://github.com/sysradium/emojify-llm/blob/main/orpo/generate.py#L26
If you run here `.save()` then with your fix it successfully produces a file `distilabel-file.json`:
```json
{
  "cuda_devices": "auto",
  "disable_cuda_device_placement": false,
  "use_magpie_template": false,
  "magpie_pre_query_template": null,
  "generation_kwargs": {
    "temperature": 0.7,
    "max_new_tokens": 256
  },
  "use_offline_batch_generation": false,
  "offline_batch_generation_block_until_done": null,
  "jobs_ids": null,
  "model": "unsloth/Llama-3.2-3B-Instruct",
  "revision": "main",
  "torch_dtype": "auto",
  "trust_remote_code": false,
  "model_kwargs": null,
  "tokenizer": null,
  "use_fast": true,
  "chat_template": null,
  "device": null,
  "device_map": null,
  "structured_output": null,
  "type_info": {
    "module": "distilabel.models.llms.huggingface.transformers",
    "name": "TransformersLLM"
  }
}
```
Without a fix raises an exception:
image

dameikle:
Thanks @sysradium. Nice ORPO example in the repos too.

dameikle:
After all that @sysradium - I'm back to where you started as the dynamic detection had a decent performance hit. I'll work my way through and marked these as excluded in the definitions. Sorry for the detour!

sysradium:
> After all that @sysradium - I'm back to where you started as the dynamic detection had a decent performance hit. I'll work my way through and marked these as excluded in the definitions. Sorry for the detour! 

I had a similar problem. However I think even with the static option performance tests were failing every other time.

dameikle:
> > After all that @sysradium - I'm back to where you started as the dynamic detection had a decent performance hit. I'll work my way through and marked these as excluded in the definitions. Sorry for the detour!
> 
> I had a similar problem. However I think even with the static option performance tests were failing every other time.

I did wonder if adding a serializer like this would be enough:
```python
    @field_serializer('*', when_used='always', return_type=Any)
    def exclude_secretstr(self, value: Any, _info) -> Any:
        """Custom serializer that excludes SecretStr fields."""
        if isinstance(value, SecretStr):
            return "********"
        return value
```

PR #1166, opened 2025-10-01
bug: quality-scorer - Improve robustness of score 

CostantinoEsposito89:
The `format_output` method in `QualityScorer` previously parsed scores by splitting the LLM output by newlines and iterating. This approach was brittle and could fail if the LLM included reasoning, explanations, or extra newlines before the score list, causing valid scores to be missed.

This commit refactors the parsing logic to use `re.findall`. This robustly extracts all lines matching the score format from the entire output, regardless of their position.

Key improvements:
- Resilience: Correctly parses scores even when they are preceded by reasoning or other text from the LLM.
- Reliability: Ensures the `scores` list is correctly padded with `None` if fewer scores are found than expected, maintaining a consistent output shape.

### DIFF
--- a/src/distilabel/steps/tasks/quality_scorer.py
+++ b/src/distilabel/steps/tasks/quality_scorer.py
@@ -187,7 +187,7 @@ class QualityScorer(Task):
         return ["scores", "model_name"]
 
     def format_output(
-            self, output: Union[str, None], input: Dict[str, Any]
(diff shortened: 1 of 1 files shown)


dameikle:
Merged in develop via squash commit with updated tests:
https://github.com/argilla-io/distilabel/commit/7dc5f70edc358fa07d05cd671af4a25de98d0850

Thanks @CostantinoEsposito89 :bow:

PR #1177, opened 2026-04-18
fix(pipeline): preserve caller parameters on GeneratorStep during dry_run

### LINKED ISSUE #1137: [BUG] Parameters of generator step are ignored during dry_run
### Describe the bug

I found that the parameters assigned to `GeneratorStep` are ignored during `dry_run`, and only `batch_size` remains. This behavior seems incorrect, as it should preserve the existing parameters and adjust only the `batch_size`.

### To reproduce

1.  Configure a pipeline with a `GeneratorStep` that includes custom parameters.
2.  Run the pipeline in `dry_run` mode.
3.  Observe that the parameters assigned to the `GeneratorStep` are lost, and only `batch_size` is retained.

### Expected behavior

During `dry_run`, the parameters assigned to `GeneratorStep` should be preserved, and only the `batch_size` should be adjusted.

### Screenshots

_No response_

### Environment

- Distilabel Version [e.g. 1.0.0]:
- Python Version [e.g. 3.11]:

### Additional context

**Actual Behavior**

The parameters assigned to `GeneratorStep` are overwritten, leaving only the `batch_size`.

**Code Snippet**

The issue seems to originate from the following code snippet:

```python
for step_name in self.dag:
    step = self.dag.get_step(step_name)[constants.STEP_ATTR_NAME]

    if step.is_generator:
        if not parameters:
            parameters = {}
        parameters[step_name] = {"batch_size": batch_size}
```

This code is located at: [https://github.com/argilla-io/distilabel/blob/main/src/distilabel/pipeline/base.py#L438](https://github.com/argilla-io/distilabel/blob/main/src/distilabel/pipeline/base.py#L438)

**Proposed Solution**

Modify the code to preserve existing parameters and only update the `batch_size`

This change ensures that existing parameters are not overwritten and only the `batch_size` is updated.

SAY-5:
### Problem

`Pipeline.dry_run` walks every generator step in the DAG and assigns
its runtime parameters with a fresh dict:

```python
for step_name in self.dag:
    step = self.dag.get_step(step_name)[constants.STEP_ATTR_NAME]
    if step.is_generator:
        if not parameters:
            parameters = {}
        parameters[step_name] = {"batch_size": batch_size}
```

That wipes whatever the caller already placed in
`parameters[step_name]` for that step, so the dry-run executes with
only `batch_size` set and loses every other runtime parameter the
user configured on the `GeneratorStep`. Dry-run results then don't
match what the full `run` produces because the step is configured
differently. Reported in #1137.

### Fix

Merge `batch_size` into the caller's dict rather than replacing it:

```diff
- parameters[step_name] = {"batch_size": batch_size}
+ parameters.setdefault(step_name, {})["batch_size"] = batch_size
```

`setdefault` preserves any existing params the caller supplied and
creates an empty dict only when they didn't set anything for that
step. Happy path (no caller params for the generator) is unchanged.

Fixes #1137


SAY-5:
Done, base is now develop and I rebased on it so the diff is just the base.py change.

